PR #36, opened 2014-06-24
Don't de-indent case and default statements

yeldarby:
This commit changes the formatting of a `switch` statement from

```
switch(x) {
case 1:
    //stuff
case 2:
    //stuff
default:
    //stuff
}
```

to

```
switch(x) {
    case 1:
        //stuff
    case 2:
        //stuff
    default:
        //stuff
}
```

Note the indentation is not decreased for `case` and `default` statements.

I know others will likely disagree with this style preference but I couldn't figure out how to control this at runtime in my init script or keymap file.

Is there a way to add this change at runtime or add a preference so I wouldn't have to maintain a fork and keep it up to date with upstream changes?

### DIFF
--- a/scoped-properties/language-javascript.cson
+++ b/scoped-properties/language-javascript.cson
@@ -13,6 +13,6 @@
       | ^ \\s* do \\s* $
       '
     'decreaseIndentPattern': '(?x)
-        ^(.*\\*/)?\\s*(\\}|\\)|case|default)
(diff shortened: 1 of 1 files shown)


kyriacos:
I know this is closed, just wondering if there a setting to modify it so existing projects that have been using the previous way don't have to reformat?

PR #92, opened 2015-01-17
Fixes #77

### LINKED ISSUE #77: Over-enthusiastic outdent of curly braces
A gif probably explains it best:

outdent2

Firstly, I'd contend that the `])` shouldn't be automatically indented, but I can live with that.

Main point: the `{whatever}` line automatically outdents as soon as I type the first `{` (presumably because it thinks it's opening a block rather than an object), and then continually tries to outdent whenever I type any more characters on the line (the re-indentation in the gif is me hitting `ctrl+]`)

Been happening in `atom --safe` built off master for the last couple of days, presumably since the update to language-javascript@0.51 in relation to e38f1cd9eddea7cb9994342ca9bceef2a4e3d7ff

UltimateBrent:
Reverts part of https://github.com/atom/language-javascript/commit/e38f1cd9eddea7cb9994342ca9bceef2a4e3d7ff that was causing overzealous decrease of indents, especially on lists of objects inside an array. Each `{` was causing a decrease of indent.


50Wliu:
With this patch, does the following still get auto-indented correctly?
```javascript
if(true)
{
    console.log('Q');
}
```

I think that line was there to prevent this from happening:
```javascript
if(true)
    {
        console.log('D:');
    }
```

UltimateBrent:
If that's the desired effect, then `{` should be removed from the increaseIndentPattern, not put in the decreaseIndentPattern. This patch merely fixes the bug, it does not address the Allman braces issue that was the original intent of https://github.com/atom/language-javascript/commit/e38f1cd9eddea7cb9994342ca9bceef2a4e3d7ff . I can look into that if you'd like to have that functionality before merging.

maxbrunsfeld:
Fixes #5180.

maxbrunsfeld:
@UltimateBrent thanks for tracking this down. @kevinsawicki what do you think we should do here?

My inclination would be to merge this PR and similarly remove the [`while|for`](https://github.com/atom/language-javascript/blob/master/settings/language-javascript.cson#L12) and [`do`](https://github.com/atom/language-javascript/blob/master/settings/language-javascript.cson#L13) `increaseIndentRegexps`. That way, both Allman and non-Allman styles would always work. It would just mean that, like @UltimateBrent said, you'd have to hit `tab` for all single-line non-bracketed control-flow statements.

kevinsawicki:
Yeah, I think merging this is fine.

PR #94, opened 2015-01-21
Add `.jsx` files to JavaScript grammars

danielbayley:


kevinsawicki:
This was explicitly removed in https://github.com/atom/language-javascript/pull/30 so that the react package would be the default to open this extension.

danielbayley:
Ah ok. That react package doesn't show up under Community Packages within Atom mind you.

PR #96, opened 2015-01-25
Autoindent rules for arrays

rclark:
As a follow-up to #77, where @jbt commented:

>  I'd contend that the ]) shouldn't be automatically indented

This adds autoindent rules for arrays, like:

```js
var pets = [
  'kitten',
  'fish'
];
```

If not auto-indenting arrays was a conscious decision, please feel free to close!


maxbrunsfeld:
Thanks!

PR #99, opened 2015-02-04
Highlighting properties of object

AMalininHere:
This change should make code more readable.


grammars/javascript.cson
```diff
@@ -363,6 +363,10 @@
     'name': 'support.constant.dom.js'
   }
   {
+    'match': '(?<=\\.)([a-zA-Z_?.$][\\w?.$]*)\\b'
```
kevinsawicki: I'm a little unclear why `?` is in these character classes, I didn't think you could have question marks in property names.

kevinsawicki:
Would you mind adding some tests for this? There is a spec folder where can see how the tests are structured.

AMalininHere:
I wrote tests for it.

AMalininHere:
Hi @kevinsawicki. Could we discuss these changes?

PR #101, opened 2015-02-17
Match iojs shebangs

### LINKED ISSUE #100: support js syntax highlighting for #!/usr/bin/env iojs
_From @jonathanong on February 17, 2015 7:8_

which is basically like `#!/usr/bin/env node`.

is this the right place to put this issue?

_Copied from original issue: atom/atom#5577_

postcasio:
Add support for files with a shebang like `#!/usr/bin/env iojs`

Fixes #100

### DIFF
--- a/grammars/javascript.cson
+++ b/grammars/javascript.cson
@@ -5,7 +5,7 @@
   '_js'
   'es6'
 ]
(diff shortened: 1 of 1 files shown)


kevinsawicki:
Thanks for adding this :+1:

PR #108, opened 2015-03-01
Add .jsx & .jsxinc to file types

danielbayley:
For use with [ExtendScript](http://en.wikipedia.org/wiki/ExtendScript).

Signed-off-by: Daniel Bayley <daniel.bayley@me.com>

### DIFF
--- a/grammars/javascript.cson
+++ b/grammars/javascript.cson
@@ -4,6 +4,8 @@
   'htc'
   '_js'
   'es6'
+  'jsx'
+  'jsxinc'
 ]
 'firstLineMatch': '^#!.*\\b(node|iojs)'
 'name': 'JavaScript'


AMalininHere:
For JSX you can use https://atom.io/packages/react

danielbayley:
Yeah that's what I thought, but it's not the same thing. `.jsx` and `.jsxinc` files are for Adobe ExtendScript, which is just JavaScript with a few add-ons, so plain JavaScript syntax highlighting is fine without having an extra package.

If you don't want to add them to core, how would I go about adding them to my init script or something?

postcasio:
@danielbayley you can add extensions for file types using the [file-types package](https://atom.io/packages/file-types). For this you'd want to add this to your `config.cson`:

```cson
'file-types':
	'jsx': 'source.js'
```

PR #114, opened 2015-03-12
added `from` to the regex matching `keyword.control.js`

Ganonside:
added the `from` keyword, used in ES6 import statements.
```js
import foo from './bar.js';
```


kevinsawicki:
Thanks!

PR #119, opened 2015-03-27
Prioritize 2-character operators

adrianlee44:
Currently operators such as `+=` and `-=` are tokenized into 2 tokens. Changing the order of operators to match should fix this problem.


kevinsawicki:
Thanks

PR #120, opened 2015-03-28
Added var and multiple cursor selection into forin snippet

Hurtak:
old behavior

new bahaviour


kevinsawicki:
Thanks for improving this :+1:

PR #121, opened 2015-03-28
Added 3rd tab stop for getElementById snippet, removed other...

Hurtak:
getElementBy* snippets because they all had the same prefix so only the last one was functional.

### DIFF
--- a/snippets/language-javascript.cson
+++ b/snippets/language-javascript.cson
@@ -35,24 +35,24 @@
   'Anonymous Generator':
     'prefix': 'g'
     'body': 'function* ($1) {\n\t$2\n}'
-  'getElementById':
-    'prefix': 'get'
-    'body': 'getElementById(${1:\'${2:id}\'})'
   'getElementsByClassName':
     'prefix': 'get'
-    'body': 'getElementsByClassName(${1:\'${2:className}\'})'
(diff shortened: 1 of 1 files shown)


snippets/language-javascript.cson
```diff
@@ -37,16 +37,7 @@
...
-  'getElementsByClassName':
-    'prefix': 'get'
-    'body': 'getElementsByClassName(${1:\'${2:className}\'})'
-  'getElementsByName':
-    'prefix': 'get'
-    'body': 'getElementsByName(${1:\'${2:name}\'})'
-  'getElementsByTagName':
-    'prefix': 'get'
-    'body': 'getElementsByTagName(${1:\'${2:tagName}\'})'
+    'body': 'getElementById(${1:\'${2:id}\'})$3'
```
kevinsawicki: What do you think about keeping all of these and just adding new prefixes instead?

* `geti` -> `getElementById`
* `getc` -> `getElementByClassName`
* `getn` -> `getElementsByName`
* `gett` ->  `getElementsByTagName`

Hurtak: Not sure about these so inside this branch I just added 3rd tab stop to all of these snippets and moved getElementById at the end so its the one which will be selected when using 'get' prefix.

kevinsawicki:
Thanks for all the great pull requests :sunny:

PR #122, opened 2015-03-28
Unified quotes in snippets

Hurtak:
Reason: majority of snippets use single quotes in them.

Also it should probably be configurable which type of quotes will be used (not sure if there is some ticket for that).


kevinsawicki:
Thanks

PR #123, opened 2015-03-28
Removed useless statement variables from try snippet

Hurtak:
And renamed parameter in try catch snippet from variable to e

Preview (old behavior at the top, new at the bottom)
try2


kevinsawicki:
:ship:

PR #125, opened 2015-03-29
:bug: Fix tokenization for anonymous functions

caiotoon:
The following cases will be properly recognized:

```javascript
Class.method = function nonAnonymous() {}
this.method = function nonAnonymous() {}
foo: function name() {}
```

Also added test case for fixed cases.

non-anon-functions

This PR fixes https://github.com/github/linguist/issues/2284.

### DIFF
--- a/grammars/javascript.cson
+++ b/grammars/javascript.cson
@@ -65,7 +65,7 @@
     'name': 'meta.function.js'
   }
   {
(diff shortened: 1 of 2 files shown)


kevinsawicki:
Great, thanks for this :ship:

Issue #128, opened 2015-04-03
import ended by newline breaks syntax highlighting


AluisioASG:
It seems that 2902c4a breaks highlighting after an ASI-relying import:
```js
import x
let y
```
The `let` keyword isn't highlighted.  Quotes inside comments are considered to open strings.

kevinsawicki:
Could you elaborate on what you mean by "ASI-relying import"?

AluisioASG:
![image](https://cloud.githubusercontent.com/assets/1904165/7016747/261fb5ee-dcc1-11e4-80f5-606ca27a3a56.png)

kevinsawicki:
Fixed by #135

PR #135, opened 2015-04-07
Fix issue #128

AMalininHere:


kevinsawicki:
Thanks for fixing this :ship:

Issue #106, opened 2015-02-28
Async / await ?


arcanis:
Is there plan to add support for this ES7 proposal ? The main thing is that async and await are added as contextual keywords (`async` is allowed before `function`, class members and fat expressions, and `await` inside async functions).

kevinsawicki:
Yup, it would be great to support all ES6 and ES7 things in this package.

buunguyen:
+1, please add support for this.

PR #136, opened 2015-04-09
Added support of async functions

AMalininHere:
ES7 async functions. [#106](https://github.com/atom/language-javascript/issues/106)


AMalininHere:
And as I can see [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar) we can just mark `await` as keyword

kevinsawicki:
Thanks!

Issue #137, opened 2015-04-15
Many symbols without scopes


benogle:
The support for js is pretty bad. Most of the issues are in the way of parsing variable names. 

Load this example into atom. When I say something is `not tagged` I mean its scope is only `.source.js`. 

```js
import {x, y} as p from 'point';  // x, y, p not tagged; comment messed after this line
const ANSWER = 42;            // ANSWER not tagged

var one, two, three, four;    // variables not tagged
var five = 5, six = 6;

function someFunc(arg1, arg2, arg3) {
  var aVariable = 'ok';       // aVariable not tagged
  var anArray = new Array()   // anArray not tagged
  anArray.slice()             // anArray not tagged
  anArray.otherMethod()       // anArray, otherMethod not tagged
  randomGlobal.prop.omg()     // not tagged

  var anObject = {
    key1: 'val1'              // keys not tagged
    key2: 'val2'
  }
}

var funcVar = function(arg1, arg2, arg3) {
}

class Car extends Vehicle {
  constructor(speed, cost) {
    super(speed);

    var cindex = Symbol('cost');
    this[cindex] = cost;      // cindex not tagged

    this.intro = `This is a car runs at ${speed}.`;
  }
}
```

Giving these symbols correct scopes will greatly improve the support in the coming `SymbolProvider` in the soon-to-be-bundled-in-core autocomplete-plus.

coopermaruyama:
+1

Pholey:
:+1:

JamesMowery:
:+1:

I'd love to see better JS support, as I'd really like to see this made possible: https://discuss.atom.io/t/is-it-possible-to-change-the-color-of-declared-variables/16731/1

It's the **one** thing that keeps me from being totally in love with Atom.

coopermaruyama:
@JamesMowery this might be of interest to you: https://atom.io/packages/language-javascript-semantic

bobrocke:
Here are some other examples:

![open-a-terminal - atom](https://cloud.githubusercontent.com/assets/2962426/7798483/62933168-02c8-11e5-8233-ae83ebad15a8.jpg)

But there are many others. This grammar appears to be pretty young.

jonkwheeler:
![screenshot 2015-05-29 12 44 09](https://cloud.githubusercontent.com/assets/3136271/7889423/76d8891e-060d-11e5-8db2-d2ddd5fa312e.jpg)

[...]

PR #138, opened 2015-04-16
Tokenize ALL_CAPS variables

50Wliu:
Things like `MY_COOL_VAR` are now tokenized under `constant.other.js` (which could be `constant.other.global.js`, wasn't sure which one was best).  Also be aware I'm not too familiar with JS so this PR could totally be on the wrong track.

First time I managed to get a language spec I wrote to pass :smile:.

Refs #137.

### DIFF
--- a/grammars/javascript.cson
+++ b/grammars/javascript.cson
@@ -372,8 +372,8 @@
     'name': 'constant.language.null.js'
   }
   {
-		'match': '(?<!\\.)\\b([A-Z][A-Z0-9_]+)(?!\\s*:)\\b'
-		'name': 'constant.other.js'
+    'match': '(?<!\\.)\\b([A-Z][A-Z0-9_]+)(?!\\s*:)\\b'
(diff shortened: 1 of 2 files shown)


grammars/javascript.cson
```diff
@@ -372,6 +372,10 @@
     'name': 'constant.language.null.js'
   }
   {
+		'match': '(?<!\\.)\\b([A-Z][A-Z0-9_]+)(?!\\s*:)\\b'
```
mnquintana: Runaway tabs! :runner:

50Wliu: I blame the web editor :wink:

mehcode:
JavaScript has `const x = 1`; if there is a way we can mark / tag `x` as a constant (following scope correctly) that'd be nice. `ALL_CAP` variables are generally constants as well from a naming perspective so I'd be okay with this too.

kevinsawicki:
Looks good, thanks!

PR #140, opened 2015-04-23
Tokenize variables declared using const as constants

50Wliu:
The `myCoolVar` in `const myCoolVar = 42` will now be tokenized as a constant.
Unfortunately this doesn't affect tokenization after the declaration...
```js
const myCoolVar = 42;  //Works!  constant.other.js
var something = 0;

something = myCoolVar * 1;  //Doesn't work :(
```
Again, refs #137.  Also, cc @kevinsawicki.

### DIFF
--- a/grammars/javascript.cson
+++ b/grammars/javascript.cson
@@ -339,7 +339,7 @@
     'name': 'storage.modifier.js'
   }
   {
-    'match': '(?<!\\.)\\b(const)(?!\\s*:)\\b\\s+(\\w+)'
+    'match': '(?<!\\.)\\b(const)(?!\\s*:)\\b\\s+([$_a-zA-Z][[$_a-zA-Z0-9]*)'
     'captures':
       '1':
         'name': 'storage.modifier.js'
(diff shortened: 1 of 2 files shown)


grammars/javascript.cson
```diff
@@ -335,10 +335,18 @@
     'name': 'storage.type.js'
   }
   {
-    'match': '(?<!\\.)\\b(const|export|extends|implements|let|private|protected|public|static|var)(?!\\s*:)\\b'
+    'match': '(?<!\\.)\\b(export|extends|implements|let|private|protected|public|static|var)(?!\\s*:)\\b'
     'name': 'storage.modifier.js'
   }
   {
+    'match': '(?<!\\.)\\b(const)(?!\\s*:)\\b\\s+(\\w+)'
```
kevinsawicki: I don't think `\w` is sufficient, it would match things like `const 5 = true;` and also not match `const $FOO = 3;`, I think it might need to be something like `([$_a-zA-Z][[$_a-zA-Z0-9]*)` instead.

50Wliu: Hmm, I totally forgot about numbers.

kevinsawicki:
Thanks for adding this :ship:

PR #148, opened 2015-05-05
Fix dollar sign prefixed support classes

### LINKED ISSUE #147: Dollar prefixed variables incorrectly partially marked as support classes
When a dollar sign is attached to a support class, it should become a standard variable. However, the support class portion of the variable name remains incorrectly marked as a support class. E.g.:

````
angular.element($window).on('storage', function() { ... });
````

This will mark "window" as a support class and leave the dollar sign unscoped.

The same happens with other support class elements, e.g. document/$document.

AdamBuczynski:
Fix #147

### DIFF
--- a/grammars/javascript.cson
+++ b/grammars/javascript.cson
@@ -411,7 +411,7 @@
     'name': 'keyword.other.js'
   }
   {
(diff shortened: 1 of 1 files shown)


kevinsawicki:
Thanks for fixing this :ship:

AdamBuczynski:
No problem, cheers.

PR #153, opened 2015-05-27
Fix not tokenizing operators in object functions

adrianlee44:
Previous, `:` was not scoped at all for object functions.
```javascript
var a = {
  hello: 'testing',
  foo: function() {},
  "testing": function() {}
}
```
The `:` in `foo: function() {}` and `"testing": function() {}` had no scope originally. This PR scopes the `:` to `keyword.operator.js` to match the scope of other `:` in objects.

### DIFF
--- a/grammars/javascript.cson
+++ b/grammars/javascript.cson
@@ -186,19 +186,21 @@
     ]
   }
   {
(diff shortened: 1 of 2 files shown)


kevinsawicki:
Thanks!

adrianlee44:
Thanks @kevinsawicki!

PR #167, opened 2015-07-07
Correct innerHeight highlighting

zephraph:
```innerHeight``` was listed as ```innerheight``` which made ```window.innerHeight``` not highlight correctly.


50Wliu:
Thanks for fixing this, @zephraph: I must have typo'd it when cleaning up the regex :grimacing:.

PR #168, opened 2015-07-09
Adding JSM file as an extension

rhelmer:
`.jsm` files are Javascript modules used by Firefox and extensions:
https://developer.mozilla.org/en-US/docs/Mozilla/JavaScript_code_modules


50Wliu:
Thanks!

PR #174, opened 2015-07-23
Add support for destructuring in const assignments

Victorystick:
I am tired of how the syntax highlighting of constant identifiers breaks when the `const` keyword is combined with destructuring assignment.

This is the current behaviour:
screen shot 2015-07-23 at 14 00 24

Behaviour as of this PR:
screen shot 2015-07-23 at 14 06 41

I would additionally like `d` to be properly highlighted as a constant above, but that will have to wait.

### DIFF
--- a/spec/javascript-spec.coffee
+++ b/spec/javascript-spec.coffee
@@ -195,6 +195,25 @@ describe "Javascript grammar", ->
       expect(tokens[6]).toEqual value: '42', scopes: ['source.js', 'constant.numeric.js']
       expect(tokens[7]).toEqual value: ';', scopes: ['source.js', 'punctuation.terminator.statement.js']
 
+      {tokens} = grammar.tokenizeLine('const {first:f,second,...rest} = obj;')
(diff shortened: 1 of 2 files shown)


spec/javascript-spec.coffee
```diff
@@ -195,6 +195,42 @@ describe "Javascript grammar", ->
       expect(tokens[6]).toEqual value: '42', scopes: ['source.js', 'constant.numeric.js']
       expect(tokens[7]).toEqual value: ';', scopes: ['source.js', 'punctuation.terminator.statement.js']
 
+      # {tokens} = grammar.tokenizeLine('const a = 1, b = a;')
```
mnquintana: Why is all of this commented out?

Victorystick: Sorry, I thought I excluded this. It was a test for multiple const declarations that I couldn't get working at this time. I'll amend the commit.

50Wliu:
Thanks!

Issue #176, opened 2015-07-25
Declaraing a const regex variable, destroys regex highlighting


Victorystick:
See example:
<img width="198" alt="screen shot 2015-07-25 at 15 05 54" src="https://cloud.githubusercontent.com/assets/703602/8889615/af6caea8-32e1-11e5-83b6-66aa9503854c.png">

50Wliu:
Fixed by #177.

PR #177, opened 2015-07-25
Fix highlighting issue when declaraing a const regex variable

Victorystick:
My new `const` matchers broke regex highlighting (and possibly others). This PR addresses #176.

Before:
screen shot 2015-07-25 at 15 05 54
After
screen shot 2015-07-25 at 15 35 47


50Wliu:
:+1:

PR #178, opened 2015-07-25
Properly escape the setTimeout snippet

### LINKED ISSUE #22: setTimeout snippet tab stop function selection problem.
The setTimeout snippet (`timeout`) attempts to select the anonymous function as a tab stop, but the closing curly on the function body closes the selection.

```
setTimeout(${2:function () {\n\t$3\n}}, ${1:10});
```

It expands to this with the second tab stop selection acting like so (notice the selection does not include the `}`)

screenshot 2014-03-29 11 20 19

I tested escaping the curly brace (the way TextMate handles this), but that didn't seem to solve the problem.

50Wliu:
Fixes #22

### DIFF
--- a/snippets/language-javascript.cson
+++ b/snippets/language-javascript.cson
@@ -82,7 +82,7 @@
     'body': 'new Promise(function(resolve, reject) {\n\t$1\n});$0'
   'setInterval function':
     'prefix': 'interval'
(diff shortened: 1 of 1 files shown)


lee-dohm:
:+1: Looks good to me!

PR #179, opened 2015-07-26
Give priority to strings in functions

### LINKED ISSUE #175: The string "){" as a function argument breaks subsequent tokens.
screen shot 2015-07-25 at 11 53 13

Passing the string `"){"` as a function argument breaks subsequent tokens. What appears to be happening is that the grammar ends the function call early at the closing bracket, not recognizing the first `"` as the start of a string, but rather the second one. Which leads to the rest of the file being understood as a string.

/ref TypeStrong/atom-typescript#481

50Wliu:
This might be a naive change, but it's the first thing I tried and it worked :P.

Fixes #175

### DIFF
--- a/grammars/javascript.cson
+++ b/grammars/javascript.cson
@@ -626,6 +626,9 @@
           {
             'include': '#function-params'
           }
+          {
+            'include': '#strings'
+          }
         ]
 
         'end': '(\\))'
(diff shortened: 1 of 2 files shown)


lee-dohm:
Looks good to me :shipit:

thomasjo:
:+1: Haven't tested this, but it looks good to me as well.

PR #181, opened 2015-07-30
Make =, of, in and ; end a const declaration.

### LINKED ISSUE #180: const in brackets causes syntax highlighting issues
When writing a `const` variable (inside some brackets), atom continues to think all the characters afterwards are still part of that variable's name, until it reaches some kind of marker, usually an `=`.

This seems to be a new bug introduced in Atom 1.0.3 as I had not experienced it prior, using the same code.

Examples of problematic code:
```js
for (const myVar of myArray) {
  //do something
}
```
```js
if (const myVar === undefined) 
  let str = 'I am the wrong color!';
```
```js
(const myVar);
window.alert("Atom still think's I'm a const!");
```

This works fine:
```js
const myVar;
//atom knows I'm no longer a const ♥

Victorystick:
Made a const declaration end with one of `=, of, in, ;`. 

Fixes #180

Before:
screen shot 2015-07-30 at 10 45 13

After:
screen shot 2015-07-30 at 10 45 35

### DIFF
--- a/grammars/javascript.cson
+++ b/grammars/javascript.cson
@@ -367,10 +367,12 @@
     'beginCaptures':
       '1':
         'name': 'storage.modifier.js'
-    'end': '(=)'
+    'end': '(=|of|in)|(;)'
     'endCaptures':
       '1':
         'name': 'keyword.operator.js'
+      '2':
+        'name': 'punctuation.terminator.statement.js'
     'patterns': [
       {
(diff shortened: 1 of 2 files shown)


50Wliu:
Ah, you got to it before me :)
I'll review this tonight.

50Wliu:
Looks good to me except for one thing which I'll PR in after this (want to get this out as quickly as possible).

That being said, are there any other possibilities where we might be too over-eager with this new const regex?

PR #182, opened 2015-07-31
More const fixes

50Wliu:
Continuation of #181.

I also removed some unneeded whitespace matchers so that `const` gets highlighted the moment you finish typing it, not after you add a space (this is consistent with all the other regex rules).  There was also an orphaned whitespace matcher without a name, so I :fire:'d that as well.

/cc @Victorystick


50Wliu:
Merging this to make sure the regressions go away in 1.0.4 - the regex can be further improved later on.

PR #184, opened 2015-08-01
Improve function-params regex

### LINKED ISSUE #183: JavaScript is broken
Hi. JavaScript the syntax highlighting is broken and this is affecting me in both Atom and GitHub code viewer. Please see the attached images.

untitled 2
untitled

https://github.com/github/linguist/issues/2530

50Wliu:
The `function-params` regex was very archaic and hard to understand, so I've replaced it with a system that doesn't have nested patterns and can be easily expanded upon.  It's also much more robust than the previous system in that it tokenizes brackets, strings, numbers, and more.

In addition, I've also added a new regex rule called `entity.name.constructor.js` for tokenizing `constructor()`-type functions.  The name can probably be improved :).

Fixes #183.


50Wliu:
@bucaran this should fix both your issues - I added in your constructor as a new spec :smile:.

/cc @Victorystick if you want to take a look.

bucaran:
:clap: Awesome.

grammars/javascript.cson
```diff
@@ -555,6 +571,13 @@
   }
 ]
 'repository':
+  'numbers':
+    'patterns': [
+      {
+        'match': '\\b((0(x|X)[0-9a-fA-F]+)|(0(b|B)[01]+)|(0(o|O)[0-7]+)|([0-9]+(\\.[0-9]+)?)((e|E)[+-]?[0-9]+)?)\\b'
```
lee-dohm: Can JavaScript decimal literals have an optional sign on the front of them too? Or is it like in Python where there are only positive literals that get negated?

lee-dohm:
Looks good to me :boat:

PR #190, opened 2015-08-05
Added "async" keyword

nkt:


50Wliu:
`async` should already be tokenized correctly - can you please share an example where it isn't?  Thanks.

nkt:
Actually it doesn't highlight as a keyword like await.
2015-08-06 03-47-10 untitled - users nkt projects pochin api - atom

mehcode:
@50Wliu It is also not highlighted inside of a class

```js
class Foo {
  async bar() {
  }
}
```

50Wliu:
Ok, thanks.  Looks like it's only supported in a couple of places.

50Wliu:
So everywhere where async is supported (such as `async function hi()`), it's tokenized as a `storage.modifier.js`.  Do you have any opinions on whether it should stay as `storage.modifier` or change everything to `keyword.operator`?

Victorystick:
What is the convention (TextMate/Atom)? One could be strict about all keywords being marked with the `keyword` scope, but on the other hand that might make for annoying highlighting, and `storage.modifier` may improve readability.

kevinsawicki:
> Do you have any opinions on whether it should stay as storage.modifier or change everything to keyword.operator?

I think using `storage.modifier` for this case to start is the right move so that all instances are consistent.

I'm going to merge this PR as is and just tweak it on master, thanks for this change :+1:

nkt:
@kevinsawicki great! Can you publish new release?

kevinsawicki:
> Can you publish new release?

I published a new version of this package, 0.88.0, which will be included in the next Atom release, 0.108.0

nkt:
:+1:

PR #192, opened 2015-08-07
Prioritize strings in object function regex

### LINKED ISSUE #76: Regular expression for function has problems
In  https://github.com/atom/language-javascript/blob/master/grammars/javascript.cson
There seems to be a problem with a particular regular expression.

If you try to code the line
```js
a.push('x' + y + ':function()');
```

I think it matches the pattern
```cson
'begin': '(?:((\')(.*?)(\'))|((")(.*?)(")))\\s*:\\s*\\b(function\\*?)?\\s*(\\*?)(\\()'
```

because as I understand it, this has a quoted string on the line  (which uses `.*?` to match the second apostrophe) using the first and third ones as the markers for the string. then there is a colon and the word "function" followed by some stuff and then a "("

I don't know this particular regular expression language well enough to change `(\')(.*?)(\'))` to exclude the apostrophe being matched by the `(.*?)`

What regular expression library defines this pattern? Is it a Ruby thing?  
And what is a cson file anyway? It reminds me of json.

50Wliu:
Previously, the `.*?` regex would match over quotes without stopping, which resulted in #76.
This PR changes that regex to `[^"]*?` (and `\'`, respectively) so that quotes will always get tokenized properly.

Fixes #76


spec/javascript-spec.coffee
```diff
@@ -398,6 +398,20 @@ describe "Javascript grammar", ->
     expect(tokens[5]).toEqual value: ')', scopes: ['source.js', 'meta.method.js', 'punctuation.definition.parameters.end.js']
     expect(tokens[6]).toEqual value: ';', scopes: ['source.js', 'punctuation.terminator.statement.js']
 
+    {tokens} = grammar.tokenizeLine('a.push("x" + y + ":function()");')
```
kevinsawicki: Mind covering the single quote case here too?

50Wliu:
@kevinsawicki done :green_apple: :white_check_mark:

kevinsawicki:
:+1:

PR #200, opened 2015-08-19
Fix highlighting bug in the 'for in' snippet

drew-y:
The closing bracket of the $3 tabstop wasn't highlighted properly on tab.


50Wliu:
Thanks!

PR #202, opened 2015-08-19
Fix bracket highlighting in the 'try' snippet

drew-y:
It looks like there was some confusion about the `\\` at 8d915f14914fd39a6a2c4896562e3dae8c94b500 (`setInterval` and `setTimeout` has already been fixed).


kevinsawicki:
Awesome, thanks so much for fixing this :ship:

Issue #191, opened 2015-08-05
`constructor` is no longer highlighted


mnquintana:
Although `constructor` is tokenized correctly, it now has no highlighting (in One Dark Syntax) as of Atom 1.0.4.

#### Example

```js
class Overlay {
  constructor(map) {
    this.map = map;
  }

  getMap() {
    return this.map;
  }
}
```

My guess is that this has something to do with https://github.com/atom/language-javascript/pull/184.

/cc @50Wliu

50Wliu:
This is because one-dark doesn't generically color `.entity` (which would cover `.entity.name.constructor` and `.entity.name.function`, which is also used a lot), only `.entity.name.type` and `.entity.other.inherited-class`.
/cc @simurai

simurai:
Coloring only `.entity` would affect other places.. for example in CSS the `::` in `::before`.

We could add `.entity.name.constructor`?

1. Should it look the same as the `getMap()` function?
2. Do other languages also define a `.entity.name.constructor` or is that only in JavaScript?

mnquintana:
Maybe it could also have `.entity.name.function` too?

50Wliu:
I'm definitely open to a name change :smile:.  Don't have much experience with how to properly name these things.

Victorystick:
How do languages like Java, C etc. manage their constructors? Should JS do something similar?

50Wliu:
When #146 gets updated (I'm working on that right now), we'll also have `.entity.name.property`.  What about adding `.entity.name` instead of just `.entity`?

MaximSokolov:
> Perhaps **entity.name.function.constructor** would work? Its syntax pretty much like any function (or method).

:+1:

50Wliu:
Wow, I just keep forgetting to push my branch that fixes this to Github...PR'ing in a moment.

PR #206, opened 2015-08-25
Follow naming conventions

### LINKED ISSUE #201: Follow naming conventions

slinto:
Fixes #201

### DIFF
--- a/grammars/javascript.cson
+++ b/grammars/javascript.cson
@@ -375,8 +375,12 @@
     'name': 'storage.type.js'
   }
   {
-    'match': '(?<!\\.)\\b(async|export|extends|implements|let|private|protected|public|static|var)(?!\\s*:)\\b'
+    'match': '(?<!\\.)\\b(async|export|extends|implements|let|private|protected|public|static)(?!\\s*:)\\b'
     'name': 'storage.modifier.js'
+  }
+  {
+    'match': '(?<!\\.)\\b(var)(?!\\s*:)\\b'
+    'name': 'storage.type.var.js'
   }
   {
     'begin': '(?<!\\.)\\b(const)(?!\\s*:)\\b'
--- a/grammars/javascript.cson
+++ b/grammars/javascript.cson
@@ -375,11 +375,11 @@
     'name': 'storage.type.js'
   }
   {
-    'match': '(?<!\\.)\\b(async|export|extends|implements|let|private|protected|public|static)(?!\\s*:)\\b'
+    'match': '(?<!\\.)\\b(async|export|extends|implements|private|protected|public|static)(?!\\s*:)\\b'
     'name': 'storage.modifier.js'
   }
   {
-    'match': '(?<!\\.)\\b(var)(?!\\s*:)\\b'
+    'match': '(?<!\\.)\\b(let|var)(?!\\s*:)\\b'
     'name': 'storage.type.var.js'
   }
   {
--- a/spec/javascript-spec.coffee
+++ b/spec/javascript-spec.coffee
@@ -190,7 +190,7 @@ describe "Javascript grammar", ->
   describe "constants", ->
     it "tokenizes ALL_CAPS variables as constants", ->
       {tokens} = grammar.tokenizeLine('var MY_COOL_VAR = 42;')
-      expect(tokens[0]).toEqual value: 'var', scopes: ['source.js', 'storage.modifier.js']
+      expect(tokens[0]).toEqual value: 'var', scopes: ['source.js', 'storage.type.var.js']
       expect(tokens[1]).toEqual value: ' ', scopes: ['source.js']
       expect(tokens[2]).toEqual value: 'MY_COOL_VAR', scopes: ['source.js', 'constant.other.js']
       expect(tokens[3]).toEqual value: ' ', scopes: ['source.js']
@@ -470,7 +470,7 @@ describe "Javascript grammar", ->
 
     it "tokenizes functions", ->
       {tokens} = grammar.tokenizeLine('var func = function nonAnonymous(')
-      expect(tokens[0]).toEqual value: 'var', scopes: ['source.js', 'storage.modifier.js']
+      expect(tokens[0]).toEqual value: 'var', scopes: ['source.js', 'storage.type.var.js']
       expect(tokens[2]).toEqual value: 'func', scopes: ['source.js', 'meta.function.js', 'entity.name.function.js']
       expect(tokens[4]).toEqual value: '=', scopes: ['source.js', 'meta.function.js', 'keyword.operator.js']
       expect(tokens[6]).toEqual value: 'function', scopes: ['source.js', 'meta.function.js', 'storage.type.function.js']


Victorystick:
You've moved `var`, but forgot `let`.

slinto:
There it is.

Victorystick:
And just like that, I'm sold. :+1:

Victorystick:
Just realized the scope you've added is `storage.type.var.js`, but I guess that works just the same for `let` since its value isn't constant, but variable.

50Wliu:
@simurai do you think this will cause any problems for syntax authors?  This change doesn't affect One Light/Dark, but still wanted to check with you anyway :).

MaximSokolov:
Maybe it would be to make a separate scope for `var` and `let`. 
`name` would be 'storage.type.${1}.js' but I am not sure.
> this will cause any problems for syntax authors?

```less
storage {
  &.modifier {
    color: @color1;
  }
  &.type {
    color: @color2;
  }
}
```

`var`, `let` would be @color2. So, yes. It would affect on some syntax themes.

MaximSokolov:
> since its value isn't constant, but variable
Nope. It's because it is `var`. Sorry, don't have docs with examples yet. I am working on it. TextMate convention is very unclear.

simurai:
> do you think this will cause any problems for syntax authors?

Not really for syntax authors, more for users that got used to a certain color. But from the core themes, it looks like __only__ Atom dark would be affected:

screen shot 2015-08-27 at 4 28 21 pm

So Atom dark could be adjusted to keep the same color. Or if Atom dark's intention is to show `storage.modifier` differently than `storage.type`, then you could argue that it's currently wrong and this change would correct it.

MaximSokolov:
> more for users that got used to a certain color

That's correct. Nothing we can do about it. At least after this PR users can change color for `var`/`let`:
```less
storage {
  &.type {
    color: @color1;
    &.var, &.let {
       color: @color2;
    }
  }
}
```

50Wliu:
Ok, I'll merge this then and see how it goes.

simurai:
> Ok, I'll merge this then and see how it goes.

:+1: In case we get issues for Atom dark about it, we can change the color back to blue for `var/let`.

MaximSokolov:
`let` should be `storage.type.let.js`. Unique scope for each language symbol.

> In case we get issues for Atom dark about it, we can change the color back to blue for var/let.

Do you want change each theme where colors for `storage.type` and `storage.modifier ` are different?

50Wliu:
> let should be storage.type.let.js

I disagree with this.  `var` and `let` are fundamentally the same.

pchaigno:
> let should be storage.type.let.js. Unique scope for each language symbol.

While I agree that having more subscopes increases customization possibilities, I think having a unique scope for each symbol is excessive. You should consider the impact it would have on the maintainability of the syntax definition.

MaximSokolov:
> I disagree with this. var and let are fundamentally the same.

If you think these scopes are fundamentally the same, you can suggest new common subscope to atom/atom#8430 like `storage.type.{common}.var` These keywords have different behavior. Is `int` and `double` are fundamentally the same? Maybe. I saw scope `primitive` in Jave. Either a new subscope or 
`storage.type.let`.

> You should consider the impact it would have on the maintainability of the syntax definition.

Don't understand. What do you mean by "maintainability of the syntax definition." Is `&.var, &.let` instead of `&.var` a big impact?

MaximSokolov:
When I see `{roorScope}.{languageSymbol}` I expected it would be match only `languageSymbol` but not `languageSymbol` and `anythingElse`

50Wliu:
@MaximSokolov I would argue that `int` and `double` should be under the same scope as well.  And regarding maintainability, `&.primitive` would be much cleaner than `&.int, &.double, &.long, &.float, &...`, no?

MaximSokolov:
@50Wliu A added a `primitive` under :question: mark. Would do you think about `let` and `var`? Should they have a common scope?

50Wliu:
Yeah.  Let's keep it at `storage.type.var.js` for now to minimize changes and then take a look at things again when Atom's nearing 2.0?

MaximSokolov:
> take a look at things again when Atom's nearing 2.0

Why? These scope are **optional**. It wouldn't broke anything.

MaximSokolov:
Also It's not about `let` and `var`. It's about common approach we should write on docs. I suggest: match sa many optional scopes as you can and if symbols have something in common create a supscope for them.

50Wliu:
> It wouldn't broke anything.

It already broke language-html's specs, and as we saw previously, it does (sometimes) change the syntax colors.  You can see an effect of that here: #191

MaximSokolov:
> You can see an effect of that here: #191

I don't think it because of this PR.

>  it does (sometimes) change the syntax colors.

Yes. `var` and `let` are highlighted correctly now.

MaximSokolov:
> You can see an effect of that here: #191

It's because `constructor` is `entity.name.constructor` should be `entity.name.function.constructor`

That exactly what I am trying to fix. Languages don't follow the convention.

MaximSokolov:
You are right. Fuck the convention. Lets keep crap syntax highlighting until Atom 2.0.

50Wliu:
This PR will not be making it into Atom 1.0.8.  The scope changes broke both language-html's specs and Atom core specs (the latter of which appear to be especially difficult to update).  In the interest of getting 1.0.8 out as painlessly as possible, Atom's reverted back to the previous version.

@MaximSokolov:
There isn't exactly a "convention" yet - as you mentioned before, the Textmate docs are only guidelines, and that's all the languages have right now.  You've proposed a convention, but that hasn't been finalized.  We intend to move forwards and I think it would be best to get others' feedback on your proposals, and then allow more people to test and adapt to the changes.  Don't forget that many users have edited their `styles.less` to fit their own needs - changing scopes without warning may cause them to receive the "wrong" color or actually lose highlighting.

/cc @atom/feedback

MaximSokolov:
> The scope changes broke Atom core specs

In that case reverted back is best solution. Agree.

MaximSokolov:
> There isn't exactly a "convention" yet - as you mentioned before, the Textmate docs are only guidelines, and that's all the languages have right now.

Yes. That's the problem. Most of languages don't follow "guidelines".

> Don't forget that many users have edited their styles.less to fit their own needs - changing scopes without warning may cause them to receive the "wrong" color or actually lose highlighting.

Don't forget how many language specific rules must add syntax theme developer to make theme looks same for general languages. Otherwise syntax theme would be look different for languages don't following  "guidelines".
`var` highlighted as `public` This is  "wrong" color.

It would be much harder to follow "guidelines"  if we would wait until Atom 2.0. Would be released a lot of new syntax themes and languages. Most of theme/language developers don't known about grammar convention. Even if they do, TextMate docs quite useless without examples. That means developers would copy code from general languages/themes. Would be a lot of languages don't following "guidelines". Would be a lot of "wrong" colors.

pchaigno:
@MaximSokolov Do you contribute to any syntax definition?

MaximSokolov:
@pchaigno:
> Do you contribute to any syntax definition?

Do you mean to any language-*? I only fixed a [small bug](https://github.com/atom/language-c/pull/79)

I saw most scopes of [languages from atom repo](https://github.com/atom?utf8=%E2%9C%93&query=language) when I worked on [Atom grammar "guidelines"](https://github.com/atom/atom/issues/8430) Some languages don't follow "guidelines" 

Also I known that syntax themes would look bad without a [language specific rules](https://github.com/MaximSokolov/photon-syntax/tree/master/styles/languages)

Are you trying to say that I don't undestand much about languages' grammar? Maybe. If you have   arguments against my suggestion it would be great if you will share them

pchaigno:
> Are you trying to say that I don't understand much about languages' grammar?

No, the discussion until now seem to show that you know what you're talking about :ok_hand: 
I just think that you underestimate the difficulty of maintaining a grammar were each symbol has its own scope.
<br>

> If you have arguments against my suggestion it would be great if you will share them

I already did ;)
> the impact it would have on the maintainability of the syntax definition.

A grammar with a scope per keyword would be really verbose and it's already difficult to debug issues in current state. I'm all for well-defined scopes but I don't think we should sacrifice the syntax definition for the possibility to customize everything.

simurai:
Would the specs pass if `var` and `let` got changed from `storage.modifier.js` to `storage.type.js`? So it would change to an existing sub-category, but not create a new one.

It would also match how TextMate currently highlights `var`s: https://tmtheme-editor.herokuapp.com/#!/editor/theme/Monokai

screen shot 2015-08-28 at 9 54 15 am

50Wliu:
No, both language-html and Atom currently test to make sure `var` is `storage.modifier.js`.

Victorystick:
That's too bad. Why is it much more difficult to update the specs of Atom core?

I agree whole-heatedly that common subscopes and adherence to a common set of guidelines is the _only_ way to make syntax themes work just as well across all languages.

I encouraged similar changes to atom/language-go and kevinsawiki/monokai. Go had a unique set of scopes that no other language used. The syntax theme had accordingly adopted a set of special rules to make Go highlight like other languages. The solution: common subscopes.

MaximSokolov:
> I just think that you underestimate the difficulty of maintaining a grammar were each symbol has its own scope.
> ...
> A grammar with a scope per keyword would be really verbose and it's already difficult to debug issues in current state. I'm all for well-defined scopes but I don't think we should sacrifice the syntax definition for the possibility to customize everything.

I undestand that it's difficult . That's why I called these scopes **optional**.

> When I see `{roorScope}.{languageSymbol}` I expected it would be match only `languageSymbol` but not `languageSymbo`l and `anythingElse`

It should be just `storage.type` then

> No, both language-html and Atom currently test to make sure var is scoped as storage.modifier.js.

Why unit tests for language-html and Atom *currently test to make sure var is scoped as storage.modifier.js*? Seems like wrong usage of modular tests.

We should discuss important things:

> changing scopes without warning may cause them to receive the "wrong" color or actually lose highlighting.

50Wliu:
> Why do unit tests for language-html and Atom currently test to make sure var is scoped as storage.modifier.js? Seems like wrong usage of modular tests.

I agree that those aren't the prettiest tests, but language-html tests to make sure that embedded JS works correctly (hence it needs to test for the JS scopes), and Atom needs to make sure grammars work, so it arbitrarily decided to use JS (and I think HTML in some other spots).

Regarding the 'lose highlighting' part:
Let's say I'm new to CSS/LESS and don't know that `&` exists.  Therefore, to target `storage.modifier.js` in my less file, I might just write
```less
.storage.modifier.js
{
    color: ...
}
```

So if `var` is suddenly changed to `storage.type.var.js`, then no more highlighting :(.  Not that common, but it could definitely happen.

MaximSokolov:
> Therefore, to target storage.modifier.js in my less file, I might just write

Good example. Beginners have to edit their less files because of wrong color of `var` scope. And changing `storage.modifier.js` would also change colors of keywords like `async`

Your example is synthetic. I agree with  `@simurai`'s supposition:
>> do you think this will cause any problems for syntax authors?

> Not really for syntax authors, more for users that **got used to a certain color**.

I don't see any way to make languages follow guidelines and don't break syntax highlighting. Still I think it worth it.

lee-dohm:
> I don't see any way to make languages follow guidelines and don't break syntax highlighting. Still I think it worth it.

I think we all agree that it is worth it. Some of us just feel that the change has to be communicated (which takes time) so that people don't feel like we've pulled the rug out from under them or that Atom is broken.

MaximSokolov:
Still no suggestion. Just wait until Atom 2.0?

lee-dohm:
That's not at all what I said. If you're intent on assuming the worst, there isn't anything I can do to dissuade you.

MaximSokolov:
> That's not at all what I said. If you're intent on assuming the worst, there isn't anything I can do to dissuade you.

Sorry for misunderstanding. I just want to known your opinion about what we have to do.
Should languages follow guidelines? Or we should stay all as it is now. And if they should, when we should change "wrong" scopes. How to notify users? Create a doc page? Should we update languages separatly? We can create a branch for each language and update all languages at the same time. No one didn't suggest anything.

lee-dohm:
Grammars and themes should both work from a well-designed, unified system. As for the rest of it, I'm working on some ideas. When I have more that I can share, I'll communicate it out.

MaximSokolov:
> > Should languages follow guidelines?

> Grammars and themes should both work from a well-designed, unified system.

I guess it means yes.

lee-dohm:
I mean what I said: Grammars and themes should both work from a well-designed, unified system. If I had meant "yes", I would have said that.

simurai:
> Just wait until Atom `2.0`?

- I think for "small fixes" that don't have a big impact: __No__, can be done already now.
- Enforcing a new naming convention/specification: __Yes__, unfortunately needs to wait for `2.0`.

With "small fixes" I mean:
- changing things that are currently wrong
- don't need changes in all themes or lots of language packages
- shouldn't annoy too many users

So I think this PR still goes under the "small fixes", because most themes are not affected. Some are, like Monokai which is pretty popular, but if TextMate's version is the original, then Atom's port has been "wrong" from the beginning and now would get "fixed". Some users might get annoyed, but some might appreciate it.

Having a new naming convention/specification before `2.0` would be kinda hard to enforce, no? We would've to:
- Update all language packages
- Update all themes
- Update all `styles.less`

Or communicate it to package/theme authors, but I think that's a hard sell trying to do it between major versions.

How to go forward?

1. Keep making "small fixes".
2. Since defining a new naming convention/specification takes time, it's great to already start thinking about it now. Even if it doesn't get added to Atom right away. After an initial draft and feedback, I like the suggestion of making a branch or fork and start to test it out with a few languages and themes. And keep refining the convention/specification until we feel confident about it.

@MaximSokolov Your enthusiasm trying to improve the grammars and syntax highlighting is great. :clap: But I hope you agree that the impact of making big changes needs to be considered too.

MaximSokolov:
Thank you, @simurai That's **exactly** what I wanted to know. Not just "blah-blah-blah... well-designed, unified system...blah-blah-blah". I hope @lee-dohm's perfect system would be in Atom 2.0.

Well, if you will look at my [proposed convention](https://github.com/atom/docs/pull/114) you will see that it differs  from the TextMate convention just little. It has **only one** new root scope — `punctuation` And it's already used in most languages. The rest of the scopes are **optional** (e.g. `variable` -> `variable.argument`) That means it wouldn't change highlighting (not sure about specs though).

>> Just wait until Atom 2.0?

> I think for "small fixes" that don't have a big impact: No, can be done already now.

I have waited for this comment. I just don't want to do ton of work for nothing.

So what I am going to do:

1. Finish work on [documentation](https://github.com/atom/docs/pull/114)

  I need your help: we should decide what scopes would be in convention (pay attention to scopes under :new: and :question: ). Also you can suggest new scopes.

2. Then I create an issue for each languages from atom repo with list of "wrong" scopes. And if changes would break syntax highlighting I add `:someBadEmoji:`

  **`:badEmoji:`** var should be storage.type.var Now: storage.modifier

MaximSokolov:
Even if we wouldn't change anything we still should have a specification. Many languages' symbols not tagged now. At least we won't have to change new symbols' scopes with clear documentation.

PR #207, opened 2015-08-25
storage.type.js scope for var

### LINKED ISSUE #201: Follow naming conventions

pchaigno:
Fixes #201.


mnquintana:
@pchaigno It looks like https://github.com/atom/language-javascript/pull/206 is trying to do the same thing

pchaigno:
@mnquintana Hm... #206 is better. I should have paid more attention to the references in the original issue.

Issue #58, opened 2014-10-26
Atom not handling multi-line strings correctly


Cydrobolt:
In Javascript, each line in a multi-line string must be ended with a `\`. Atom does not recognise this need, and highlights the whole string rather than warning users about missing `\`s.

oferh:
+1

rossPatton:
wouldn't be something that your linter would catch? i know eslint yells at you for this

50Wliu:
Fixed by #209.

PR #209, opened 2015-08-26
Highlight unescaped multiline strings as invalid

pchaigno:
This pull request implements a change requested in #58.
Currently unescaped multiline strings are highlighted as if they were correct:
```js
'line1
line2\
line3'
```
This change will highlight such strings as invalid.
[Example of an invalid string with this change in Lightshow](https://github-lightshow.herokuapp.com/?utf8=%E2%9C%93&scope=from-url&grammar_url=https%3A%2F%2Fraw.githubusercontent.com%2Fpchaigno%2Flanguage-javascript%2Fa906d863677b8346aba3d429434a63747ac334ba%2Fgrammars%2Fjavascript.cson&grammar_text=&code_source=from-text&code_url=&code=%27line1%0D%0Aline2\\%0D%0Aline3%27).

/cc @Cydrobolt

### DIFF
--- a/spec/javascript-spec.coffee
+++ b/spec/javascript-spec.coffee
@@ -31,6 +31,21 @@ describe "Javascript grammar", ->
         expect(tokens[2].value).toEqual delim
         expect(tokens[2].scopes).toEqual ["source.js", scope, "punctuation.definition.string.end.js"]
 
+    it "tokenizes invalid multiline strings", ->
+      lines = grammar.tokenizeLines("'line1\nline2\\\nline3'")
+      expect(lines[0][0]).toEqual value: "'", scopes: ['source.js', 'string.quoted.single.js', 'punctuation.definition.string.begin.js']
+      expect(lines[0][1]).toEqual value: 'line1', scopes: ['source.js', 'string.quoted.single.js', 'invalid.illegal.string.js']
+      expect(lines[1][0]).toEqual value: 'line2\\', scopes: ['source.js', 'string.quoted.single.js']
+      expect(lines[2][0]).toEqual value: 'line3', scopes: ['source.js', 'string.quoted.single.js']
+      expect(lines[2][1]).toEqual value: "'", scopes: ['source.js', 'string.quoted.single.js', 'punctuation.definition.string.end.js']
+
+      lines = grammar.tokenizeLines('"line1\nline2\\\nline3"')
+      expect(lines[0][0]).toEqual value: '"', scopes: ['source.js', 'string.quoted.double.js', 'punctuation.definition.string.begin.js']
+      expect(lines[0][1]).toEqual value: 'line1', scopes: ['source.js', 'string.quoted.double.js', 'invalid.illegal.string.js']
+      expect(lines[1][0]).toEqual value: 'line2\\', scopes: ['source.js', 'string.quoted.double.js']
+      expect(lines[2][0]).toEqual value: 'line3', scopes: ['source.js', 'string.quoted.double.js']
+      expect(lines[2][1]).toEqual value: '"', scopes: ['source.js', 'string.quoted.double.js', 'punctuation.definition.string.end.js']
+
   describe "keywords", ->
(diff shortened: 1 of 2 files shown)


spec/javascript-spec.coffee
```diff
@@ -31,6 +31,21 @@ describe "Javascript grammar", ->
         expect(tokens[2].value).toEqual delim
         expect(tokens[2].scopes).toEqual ["source.js", scope, "punctuation.definition.string.end.js"]
 
+    it "tokenizes invalid multiline strings", ->
+      lines = grammar.tokenizeLines("'line1\nline2\\\nline3'")
```
50Wliu: It would be nice if you can span this over multiple lines instead of using `\n` (see examples further down)

spec/javascript-spec.coffee
```diff
@@ -31,6 +31,21 @@ describe "Javascript grammar", ->
...
+      expect(lines[0][0]).toEqual value: "'", scopes: ['source.js', 'string.quoted.single.js', 'punctuation.definition.string.begin.js']
+      expect(lines[0][1]).toEqual value: 'line1', scopes: ['source.js', 'string.quoted.single.js', 'invalid.illegal.string.js']
+      expect(lines[1][0]).toEqual value: 'line2\\', scopes: ['source.js', 'string.quoted.single.js']
+      expect(lines[2][0]).toEqual value: 'line3', scopes: ['source.js', 'string.quoted.single.js']
+      expect(lines[2][1]).toEqual value: "'", scopes: ['source.js', 'string.quoted.single.js', 'punctuation.definition.string.end.js']
+
+      lines = grammar.tokenizeLines('"line1\nline2\\\nline3"')
```
50Wliu: This test isn't needed if you loop the above test with a for loop as in the first string test (`delimsByScope`)

spec/javascript-spec.coffee
```diff
@@ -31,6 +31,21 @@ describe "Javascript grammar", ->
...
+      expect(lines[0][0]).toEqual value: '"', scopes: ['source.js', 'string.quoted.double.js', 'punctuation.definition.string.begin.js']
+      expect(lines[0][1]).toEqual value: 'line1', scopes: ['source.js', 'string.quoted.double.js', 'invalid.illegal.string.js']
+      expect(lines[1][0]).toEqual value: 'line2\\', scopes: ['source.js', 'string.quoted.double.js']
+      expect(lines[2][0]).toEqual value: 'line3', scopes: ['source.js', 'string.quoted.double.js']
+      expect(lines[2][1]).toEqual value: '"', scopes: ['source.js', 'string.quoted.double.js', 'punctuation.definition.string.end.js']
+
```
50Wliu: Can you also please add a test illustrating a correct multiline string?

50Wliu:
Thanks a lot for this @pchaigno, tried solving this by myself a few weeks ago but couldn't get it to work.
Left some small feedback, and then it should be good to go.

spec/javascript-spec.coffee
pchaigno: Done.

spec/javascript-spec.coffee
pchaigno: Ooh, neat!
Done in 99d50c2

spec/javascript-spec.coffee
pchaigno: I built this example so that `line2` is valid. Do you think I should break that down into two tests?

pchaigno:
> Thanks a lot for this @pchaigno, tried solving this by myself a few weeks ago but couldn't get it to work.

You're welcome. That was a tough one.
And there seem to be some differences in the way Atom and GitHub handle lookarounds. I'll check the [implementation for GitHub](https://github.com/vmg/pcre) later.

50Wliu:
:boat:

pchaigno:
:tada:

Cydrobolt:
:+1:

PR #146, opened 2015-04-30
Added support for properties

AMalininHere:
Issues #37 , #97 , #137

PR #158, opened 2015-06-09
Relax support.function.js and support.constant.js regex

50Wliu:
The patterns they match are no longer a hardcoded set of words and are now just `\w+` to match any word sequence.

A side effect of this change is that there can no longer be `support.{function|constant}.dom.js` capturing groups anymore unless you wanted `.dom` to have a specific set of patterns while `support.{function|constant}` didn't.  In that sense, this is a pretty major change.

Refs #137.

PR #210, opened 2015-08-27
Add support for properties and relax support.function regex

### LINKED ISSUE #37: Hightlight functions and methods
Hello,

Currently `myObj.init().foo.bar()` and `some()` are displayed in a single color.
It would really great the functions and methods are Highlighted.

It's possible?

50Wliu:
:rotating_light: :warning: Do not merge yet!  Specs on atom/atom also need to be updated! :warning: :rotating_light:

This PR brings back #158 and combines it with #146.

* `support.function.dom.js` has been kept and is prioritized over the now-generalized `support.function.js` regex.
* `entity.name.property.js` now catches anything that doesn't fall into `support.constant{.dom}.js` and `support.class.js`

Before:
before

After:
after

/cc @bucaran since I know you had some concerns about this
/cc @nevir because this only adds `.property` for things not already marked as `.class` and `.constant`
/cc @Victorystick, @AMalininHere, @pchaigno

Refs #137
Fixes #37
Closes #146

<sub>Sorry for all the commits</sub>


jerone:
Will this detect functions executed with `.call` and `.apply` as functions? e.g.:
```
Array.prototype.shift.call(arguments);
```

50Wliu:
@jerone 
call

50Wliu:
Going to close and re-open as a different PR to try to slim down the commits (and the fact that git is refusing to sync properly).

PR #214, opened 2015-08-28
Fix / not being tokenized as an operator

### LINKED ISSUE #212: Missing scope on operator
The `/` character doesn't have a scope:

jsbug

50Wliu:
I had to move the group down because `/` started to override the regex group.

Fixes #212


Victorystick:
Simple, and good. :ship:

PR #215, opened 2015-08-28
Isolate periods in console methods

### LINKED ISSUE #213: Periods are not isolated from console methods
The period should have its own scope. Currently it is included in the method's scope:

jsbug

50Wliu:
Also adds some more support functions and adds some safeguards (eg `.log()` isn't valid anymore and neither is `console.log` - it has to be `something.log()`).

Fixes #213

### DIFF
--- a/grammars/javascript.cson
+++ b/grammars/javascript.cson
@@ -354,7 +354,7 @@
     'name': 'entity.name.type.object.js.console'
   }
   {
-    'match': '\\b(\\.)(assert|clear|debug|error|info|log|profile|profileEnd|time|timeEnd|warn)\\b(?=\\()'
+    'match': '\\b(\\.)(assert|clear|debug|error|info|log|profile|profileEnd|time|timeEnd|warn)(?=\\()'
     'captures':
       '1':
         'name': 'meta.delimiter.method.period.js'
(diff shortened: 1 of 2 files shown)


grammars/javascript.cson
```diff
@@ -354,8 +354,12 @@
     'name': 'entity.name.type.object.js.console'
   }
   {
-    'match': '\\.(warn|info|log|error|time|timeEnd|assert)\\b'
-    'name': 'support.function.js.console'
+    'match': '\\b(\\.)(assert|clear|debug|error|info|log|profile|profileEnd|time|timeEnd|warn)\\b(?=\\()'
```
pchaigno: Are the two `\b` required here?

chbk: If you want to be really accurate you could use:

    (?<=\\bconsole)\\s*(\\.)\\s*(assert|clear|debug|error|info|log|profile|profileEnd|time|timeEnd|warn)\\b

to catch:

    console.log

and:

    console
            . log ()

but not:

    console.logs

or:

    notconsole . log

50Wliu: The first one is definitely required to prevent `console,.log()` from matching.  It doesn't look like the second one is needed though; I'll remove that.

Victorystick:
There we are. :shipit:

PR #217, opened 2015-08-29
Add support for properties and relax support.function regex

### LINKED ISSUE #37: Hightlight functions and methods

50Wliu:
:rotating_light: :warning: Do not merge yet!  Specs on atom/atom also need to be updated! :warning: :rotating_light:

This PR replaces #210, brings back #158, and combines it with #146.

* `support.function.dom.js` has been kept and is prioritized over the now-generalized `support.function.js` regex.
* `entity.name.property.js` now catches anything that doesn't fall into `support.constant{.dom}.js` and `support.class.js`

Before:
before

After:
after

/cc @bucaran since I know you had some concerns about this
/cc @nevir because this only adds `.property` for things not already marked as `.class` and `.constant`
/cc @Victorystick, @AMalininHere, @pchaigno

Refs #137
Fixes #37
Closes #146

### DIFF
--- a/grammars/javascript.cson
+++ b/grammars/javascript.cson
@@ -488,12 +488,12 @@
     'name': 'support.class.js'
   }
   {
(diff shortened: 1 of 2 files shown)


MaximSokolov:
> support — things provided by a framework or library should be below support.
 - function — functions provided by the framework/library.

```js
hello.prop.anotherProp.notAProp().yesProp.window;
```
Why you use `support.function` for 'notAProp'? 'notAProp' would be  `support.function` only if 'hello' is a  `support.class` or `support.type` (I assume that we don't known if 'notAProp' is a library function) I don't think it's a good idea to use`support.function` for all methods. IMO for method calls it can be:
```js
hello.wow(arg1, arg2)
```
'wow(arg1, arg2)' — `meta.method-call`
  - 'wow' — `entity.name.function`
  - '(' — `punctuation.definition.arguments.begin`
  - 'arg1', 'arg2' — `variable.other.argument`
  - ','  — `punctuation.separator.arguments`
  - ')' — `punctuation.definition.arguments.end`

I wraped 'wow(arg1, arg2)' with `meta.method-call` because I don't think 'wow' should be the same color as  'wow' in function declaration. So it will be possible to change color for 'wow' with styles for `meta.method-call` -> `entity.name.function`

> entity — an entity refers to a larger part of the document
- name — we are naming the larger entity.
  - function — the name of a function.
  - type — the name of a type declaration or class.
  - tag — a tag name.
  - section — the name is the name of a section/heading.

Also not sure about `entity.name` for properties. Shoud it be the same color as class/funciton name since property isn't always a class or a function:
```js
a.prop = 2
```
`variable.other.property` maybe?

50Wliu:
What would you suggest notAProp() to be then?

MaximSokolov:
I wrote about it above:
`meta.method-call` for 'notAProp()' and `entity.name.function` for 'notAProp' (like in functions declarations with `meta.function`)
[Same for functions calls](https://github.com/atom/docs/pull/114#issuecomment-136132055)

PR #218, opened 2015-08-30
Scope constructor to entity.name.function.constructor

### LINKED ISSUE #191: `constructor` is no longer highlighted

50Wliu:
Closes #191

### DIFF
--- a/grammars/javascript.cson
+++ b/grammars/javascript.cson
@@ -265,7 +265,7 @@
     'begin': '\\b(constructor)\\s*(\\()'
     'beginCaptures':
       '1':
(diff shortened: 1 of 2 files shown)


Victorystick:
Wooo! :ship:

PR #219, opened 2015-09-01
Stop tokenizing $0 and $00 as replacement variables.

adamfranco:
At least in Firefox, Chrome, and the Atom Find-and-Replace, $0 and $00 are
treated as plain string values while $& is used as the variable to hold the
full match.

This commit updates the replacement grammar to only allow integers 1-99,
with and without a leading 0.

Here is the new behavior in Atom Find-and-Replace:
screen shot 2015-09-01 at 10 58 16 am

### DIFF
--- a/grammars/regular expression replacement (javascript).cson
+++ b/grammars/regular expression replacement (javascript).cson
@@ -10,7 +10,7 @@
   'regexp-replacement':
     'patterns': [
       {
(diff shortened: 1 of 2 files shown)


50Wliu:
:+1:

adamfranco:
Thanks for the quick review @50Wliu . :smiley:

Issue #188, opened 2015-08-04
Distinguishing from comparison operators


smaili:
I've noticed that these two rules use the same grouping:

```cson
  {
    'match': '(?<!\\.)\\b(delete|in|of|instanceof|new|typeof|void)(?!\\s*:)\\b'
    'name': 'keyword.operator.js'
  }
  {
    'match': '!=|!==|<=|>=|<<=|>>=|>>>=|\\*=|(?<!\\()/=|%=|\\+=|\\-=|&=|\\^=|!|%|&|\\*|\\-\\-|\\-|\\+\\+|\\+|~|===|==|=|<>|<|>|!|&&|\\|\\||\\?|\\:|\\^'
    'comment': 'match 2-character operator first'
    'name': 'keyword.operator.js'
  }
```

By using the same groups for both, we can't have syntax highlighting be applied to keywords like `delete` and `typeof`.  What would be the best way to distinguish these?  Maybe remove the `keyword` class from the second group of operators?

chbk:
How about:

- `keyword.operator.language.js` for `delete`, `in`, `of`, `instanceof`, `new`, `typeof`, `void`
- `keyword.operator.js` for `=`, `!`, `<`, `>`, `+`, `-`, `%`, and other symbols

Since the `language` class is used by `true`, `false`, `null`, `super`, `this`, `Infinity`, `NaN`, and `undefined`, it would fit nicely with the operators.

smaili:
I like it :+1:

MaximSokolov:
> Since the language class is used by true, false, null, super, this, Infinity, NaN, and undefined, it would fit nicely with the operators.

Disagree. `variable.language` scope separate user defined variables from language variables. Scopes should be semantic.  I don't see any reason why `delete, in, of, instanceof, new, typeof, void` are **language** operators and `*=, !, <, >, +, -, %, and other symbols` are not. 

 I [proposed](https://github.com/atom/docs/pull/114) `keyword.operator.comparison` , `keyword.operator.arithmetic` , `keyword.operator.logical` (and others) scopes. Seems these scopes   are suitable for this goal.

>
variable
  - parameter
  - language — **reserved language variables like this, super, self, etc.**
  - other

chbk:
> I don't see any reason why delete, in, of, instanceof, new, typeof, void are language operators and *=, !, <, >, +, -, %, and other symbols are not.

[...]

PR #222, opened 2015-09-07
Distinguish comparison operators

pchaigno:
Distinguish comparison operators as proposed in #188.


50Wliu:
@pchaigno do you want to continue this PR or move discussion over to #225?

PR #224, opened 2015-09-09
Recognize function bodies

pchaigno:
With this pull request, function and method bodies will be recognized under the scopes `meta.function.js` and `meta.method.js` with the headers.
This is a necessary step to then recognize objects.

### DIFF
--- a/grammars/javascript.cson
+++ b/grammars/javascript.cson
@@ -60,7 +60,7 @@
     'name': 'meta.class.js'
   }
   {
-    'begin': '([a-zA-Z_?.$][\\w?.$]*)\\.(prototype)\\.([a-zA-Z_?.$][\\w?.$]*)\\s*(=)\\s*(?:(async)(?:\\s+))?(function\\*?)\\s*(\\*?)(\\()'
+    'begin': '([a-zA-Z_?.$][\\w?.$]*)\\.(prototype)\\.([a-zA-Z_?.$][\\w?.$]*)\\s*(=)\\s*(?:(async)(?:\\s+))?(function\\*?)\\s*(\\*?)(?=\\()'
     'beginCaptures':
       '1':
         'name': 'support.class.js'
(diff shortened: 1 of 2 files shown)


pchaigno:
@50Wliu Travis build fixed!

MaximSokolov:
Hmm. All languages use `meta.function` scope only for function declaration. [TextMate manual](http://manual.macromates.com/en/language_grammars#naming_conventions) says that `meta.function` match function declaration:
> ... For example the entire line which declares a function would be `meta.function`...

I doubt we should change it.

I [proposed](https://github.com/atom/docs/pull/114#issuecomment-136132055) `meta.function-call` and `meta.method-call` scopes expecting that `meta.function` would match functions' declarations. So if we merge this PR this code:
```Less
.meta.function {
  .entity.name.function {
      color: @color1;
  }
}
``` 
... would also highlight functions/methods calls:
```js
function() {
  // `functionCall` would be highlighted
  // I assume `functionCall` is `entity.name.function`
  functionCall();
}
```
I suggest to add new scope under `meta.*` for functions' bodies.

50Wliu:
Can we just make `function-params` also have the => part?  Seems like a win-win.  In addition, `( x =>` doesn't seem valid to me.  `x =>` or `(x) =>` should be the only two that are valid, but I'm not exactly a Javascript expert.

MaximSokolov:
How about `declaration` sub-scope?
```less
.function.declaration {
  .entity.name {
      color: @color1;
  }
}

.function-call, .method-call {
  .entity.name {
      color: @color2;
  }
}

```

grammars/javascript.cson
```diff
@@ -76,18 +76,19 @@
...
-        'name': 'punctuation.definition.parameters.begin.js'
     'comment': 'match stuff like: Sound.prototype.play = function() { … }'
-    'end': '(\\))'
+    'end': '\\}'
     'endCaptures':
-      '1':
-        'name': 'punctuation.definition.parameters.end.js'
+      '0':
+        'name': 'punctuation.definition.function.body.end.js'
     'name': 'meta.function.prototype.js'
```
50Wliu: I think this needs to be updated, because the entire function block shouldn't be tokenized as a prototype.

50Wliu:
@pchaigno A few comments:
1. The `name`s of each have to be updated I think.  Maybe just move it to capture `0` of `beginCaptures` or give it a new name.
2. Can `function-params` be included in `function-body`?
3. There's conflicts :(

Issue #93, opened 2015-01-21
`true` and `false` don't get picked up inside ternary operator


jbt:
If you type `a ? true : true` then the first `true` doesn't get picked up as a keyword, but the second does. Same if you replace it with false

It looks to me like it's seeing the colon and interpreting the first `true` as a label or a property name or something

albertorestifo:
+1

Example:
![image](https://cloud.githubusercontent.com/assets/1037048/6412906/08efe3d6-be8a-11e4-8bcd-bec5341727ff.png)

abe33:
> It looks to me like it's seeing the colon and interpreting the first true as a label or a property name or something

By using the `Editor: Log Cursor Scope` command it appears that both `?` and `:` are marked with the `keyword.operator.js` scope but the boolean between them doesn't receive any scope.

abe33:
So here it is: 
https://github.com/atom/language-javascript/blob/master/grammars/javascript.cson#L326
https://github.com/atom/language-javascript/blob/master/grammars/javascript.cson#L331

Booleans aren't highlighted if followed by `:`. The same issue will also appears for `null`, `super`, `this`, and some node-only global like `process` or `__filename`.

I thought at first that something like below could fix the issue but it cannot, as oniguruma doesn't allow for variable-length lookbehinds (see http://stackoverflow.com/questions/3479131/problem-with-quantifiers-and-look-behind): 

```cson
match: '(?<!\\.)\\btrue(?!\\s*:)\\b|(?<=\\?\\s*)\\btrue(?=\\s*:)\\b'
```

Without the quantifier in the look-behind the regexp is valid and match `true` in `a ? true : true`. 
But having a rule for every number of space before after the `?` is just insane, so I wonder what other option is available to solve this problem? 
Except by creating a special *keywords in ternary operator* rule for each cases I listed above.

YuriSolovyov:
same for null:
![screenshot from 2015-05-25 12 38 24](https://cloud.githubusercontent.com/assets/2168518/7795182/0671a7f8-02db-11e5-83e2-c08b6f5b2a84.png)

Alhadis:
Constants aren't highlighted if followed by ternary operators, either. E.g.:

```js
var TEXT_CONTENT = "textContent";
var HTML = document.documentElement;
TEXT_CONTENT in HTML ? TEXT_CONTENT : "innerText"
```
Seems to affect GitHub's JS highlighting, too.

jbt:
:+1: nice one :)

PR #227, opened 2015-09-21
Allow regexp to immediately follow `?` ternary operator

mislav:
In the expression `a ? /b/ : /c/`, the first regexp wasn't recognized as such, but was tokenized as a sequence of operators instead. This fixes it by allowing regexp to immediately follow the `?` operator.

This fixes the syntax highlighting issue in Atom as well as on GitHub.com that manifests itself like this:

screen shot 2015-09-21 at 5 48 52 pm

This is related to #93

### DIFF
--- a/grammars/javascript.cson
+++ b/grammars/javascript.cson
@@ -553,7 +553,7 @@
     'name': 'constant.language.js'
   }
   {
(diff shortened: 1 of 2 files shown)


50Wliu:
Thanks!

mislav:
Thanks for the quick merge! :sparkles:

PR #228, opened 2015-09-21
Fix tokenizing of various keywords within ternary expression

### LINKED ISSUE #93: `true` and `false` don't get picked up inside ternary operator

mislav:
Supports proper tokenizing of these keywords when they appear as subexpression `b` within ternary expression `a ? b : c`:

    yield, true, false, null, [CONSTANT], super, this, debugger,
    module, exports, __filename, __dirname, global, process

They are not tokenized because there was a rule that these keywords should not be followed by `:`. This was to compensate for JS key/value syntax `{ foo: bar }`.

The fix adds an alternate pattern for each of these keywords that explicitly allows these keywords to be followed by `:` if they were preceded by `?`.

Fixes #93

### DIFF
--- a/grammars/javascript.cson
+++ b/grammars/javascript.cson
@@ -468,11 +468,15 @@
     ]
   }
   {
-    'match': '(?<!\\.)\\b(yield)(?!\\s*:)\\b(?:\\s*(\\*))?',
+    'match': '(?<!\\.)\\b(yield)(?!\\s*:)\\b(?:\\s*(\\*))?|(?<=\\?)(?:\\s*)(yield)(?=\\s*:)',
     'captures':
       '1':
         'name': 'keyword.control.js'
       '2':
+        'name': 'storage.modifier.js'
+      '3':
+        'name': 'keyword.control.js'
+      '4':
         'name': 'storage.modifier.js'
     'name': 'meta.control.yield.js'
   }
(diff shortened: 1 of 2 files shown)


grammars/javascript.cson
```diff
@@ -545,8 +565,13 @@
     'name': 'support.function.js'
   }
   {
-    'match': '(?<!\\.)\\b(module|exports|__filename|__dirname|global|process)(?!\\s*:)\\b'
-    'name': 'support.variable.js'
+    'match': '(?<!\\.)\\b()(?!\\s*:)\\b'
```
50Wliu: I don't think this should belong here? :P

spec/javascript-spec.coffee
```diff
@@ -53,6 +53,28 @@ describe "Javascript grammar", ->
...
+      null: 'constant.language.null.js'
+      true: 'constant.language.boolean.true.js'
+      false: 'constant.language.boolean.false.js'
+      debugger: 'keyword.other.js'
+      exports: 'support.variable.js'
+      __filename: 'support.variable.js'
+
+    for keyword, scope of map
+      do (keyword, scope) ->
+        it "does not tokenize `#{keyword}` when object key", ->
```
50Wliu: :art: `when it is an object key`

spec/javascript-spec.coffee
```diff
@@ -53,6 +53,28 @@ describe "Javascript grammar", ->
...
+          {tokens} = grammar.tokenizeLine("#{keyword}: 1")
+          expect(tokens[0]).toEqual value: keyword, scopes: ['source.js']
+          expect(tokens[1]).toEqual value: ':', scopes: ['source.js', 'keyword.operator.js']
+
+        it "tokenizes `#{keyword}` in ternary expression", ->
```
50Wliu: `when it is in a ternary expression`

spec/javascript-spec.coffee
```diff
@@ -53,6 +53,28 @@ describe "Javascript grammar", ->
       {tokens} = grammar.tokenizeLine('with')
       expect(tokens[0]).toEqual value: 'with', scopes: ['source.js', 'keyword.control.js']
 
+    map =
```
50Wliu: Can you make it so the test itself isn't looped, only the contents within the test?  So
```coffee
it "...", ->
  map =
    ...
  for keyword, scope of map
    ...
```

If necessary you can keep `map` out of the tests to avoid duplication.

spec/javascript-spec.coffee
```diff
@@ -53,6 +53,28 @@ describe "Javascript grammar", ->
...
+        it "does not tokenize `#{keyword}` when object key", ->
+          {tokens} = grammar.tokenizeLine("#{keyword}: 1")
+          expect(tokens[0]).toEqual value: keyword, scopes: ['source.js']
+          expect(tokens[1]).toEqual value: ':', scopes: ['source.js', 'keyword.operator.js']
+
+        it "tokenizes `#{keyword}` in ternary expression", ->
+          {tokens} = grammar.tokenizeLine("a ? #{keyword} : b")
+          expect(tokens[2]).toEqual value: ' ', scopes: ['source.js']
+          expect(tokens[3]).toEqual value: keyword, scopes: ['source.js', scope]
+
```
50Wliu: Just to be comprehensive can you also add a test for `"a ? b : #{keyword}"`?

spec/javascript-spec.coffee
```diff
@@ -336,6 +358,15 @@ describe "Javascript grammar", ->
       expect(tokens[4]).toEqual value: 'systemLanguage', scopes: ['source.js', 'support.constant.js']
       expect(tokens[5]).toEqual value: ';', scopes: ['source.js', 'punctuation.terminator.statement.js']
 
+    it "does not tokenize constant when object key", ->
```
50Wliu: `it "does not tokenize constants when they are object keys", ->

spec/javascript-spec.coffee
```diff
@@ -336,6 +358,15 @@ describe "Javascript grammar", ->
...
+      {tokens} = grammar.tokenizeLine('FOO: 1')
+      expect(tokens[0]).toEqual value: 'FOO', scopes: ['source.js']
+      expect(tokens[1]).toEqual value: ':', scopes: ['source.js', 'keyword.operator.js']
+
+    it "tokenizes constant in ternary expression", ->
```
50Wliu: `tokenizes constants in ternary expressions"`

spec/javascript-spec.coffee
```diff
@@ -336,6 +358,15 @@ describe "Javascript grammar", ->
...
+      {tokens} = grammar.tokenizeLine('a ? FOO : b')
+      expect(tokens[3]).toEqual value: 'FOO', scopes: ['source.js', 'constant.other.js']
+
```
50Wliu: Same thing here, a test for `'a ? b : FOO'` would be nice.

spec/javascript-spec.coffee
```diff
@@ -399,6 +430,15 @@ describe "Javascript grammar", ->
       expect(tokens[0]).toEqual value: 'yield', scopes: ['source.js', 'meta.control.yield.js', 'keyword.control.js']
       expect(tokens[2]).toEqual value: '*', scopes: ['source.js', 'meta.control.yield.js', 'storage.modifier.js']
 
+    it "does not tokenize yield when object key", ->
```
50Wliu: Same comments as above.

50Wliu:
I dislike how we have to add duplicate values for all these rules - I'm guessing you already tried to combine just the non-capturing rules and failed?

Also, it feels like we can add two more `describe`s specifically for object keys and ternaries, again to avoid duplication.

grammars/javascript.cson
mislav: Good catch, thanks

spec/javascript-spec.coffee
mislav: I like how the loop generates multiple tests because when tokenizing of one keyword breaks, only its test will fail and not the entire block. This was useful for me in development when I wrote these tests first and changed one by one regular expression to get them passing.

spec/javascript-spec.coffee
mislav: Sure, good call

mislav:
> I dislike how we have to add duplicate values for all these rules - I'm guessing you already tried to combine just the non-capturing rules and failed?

Yes, I didn't have another idea. Basically we want to disallow these keywords being tokenized as such when they're followed by `:`, but only if they're not preceded with `?`. I'm not sure if this can be described using a simpler regular expression.

mislav:
I've performed the desired changes.

spec/javascript-spec.coffee
50Wliu: > only its test will fail and not the entire block

I think you've actually convinced me to start writing new tests this way :smile:!

50Wliu:
Woo! :shaved_ice:

BrandonZacharie:
:+1:

Alhadis:
Mislav, you're awesome. If I were there I'd give ya the heartiest pat-on-the-back that I could give ya. :D :+1:

Issue #220, opened 2015-09-04
Syntax highlighting breaks on ' in JSX


rkuykendall:
I'm not sure if this is the right repo, but from what I can tell this is what Github uses for syntax highlighting. This code looks fine in my Atom, but breaks on github at a ' in JSX.

```javascript
'use strict';

const React = require('react')

module.exports = React.createClass({
  render: function() {
    let {feeds, log} = this.props;

    log.info(feeds);
    return <div className="feed-list">
      <h3>News Feed's</h3>
      <ul>
        {feeds.map(function(feed) {
          return <li key={feed.name} className={feed.fetched ? 'loaded' : 'loading'}>
            {feed.data && feed.data.length > 0 ?
              <span>{feed.name} <span className='light'>({feed.data.length})</span></span>
              : 'feed.name' }
          </li>
        })}
      </ul>
    </div>;
  }
});
```

Victorystick:
What you're seeing is language-javascript not supporting JSX at all...

`<h3>` is tokenized as `operator, Identifier, operator`. It doesn't know JSX considers it an opening tag. Once `'` is reached out tries matching what it considers a string literal. However, the literal isn't closed before the newline, which is a syntax error.

rkuykendall:
Do you know why this is broken on Github snippets but not the Atom editor? Are they not using the same repos?

Victorystick:
They _are_ both using the same repo, but they're not always using the exact same version. `language-javascript` is changing pretty fast, so it's hard to keep up.

Victorystick:
Are you running the latest version of Atom? It was only released two days ago, and I'm pretty sure it includes this particular behavior for strings.

50Wliu:
@rkuykendall what file extension does your file have?  .jsx or .js?

rkuykendall:
@50Wliu The original was .js, but I made a Gist with .es6 and .jsx and got the same thing.

@Victorystick Ah, I just double-checked my Atom and it looks like I'm using the `Babel ES6 JavaScript` syntax highlighter, and switching it to `Javascript` does indeed act like this.

50Wliu:
language-javascript doesn't support JSX at this point.  @pchaigno do you know what language Github uses to parse .es6 and .jsx files?

pchaigno:
[...]

PR #230, opened 2015-10-01
add adobe extendscript extension

lmunro:


50Wliu:
Sorry, but `.jsx` is also used by [JSX](https://facebook.github.io/jsx/), which language-javascript doesn't support (see #220 for some rationale behind this).

You can add `.jsx` in your own version of Atom through something like this in your `config.cson`:
```cson
'*'
core:
  customFileTypes:
    'source.js': [
      'jsx'
    ]
```

PR #236, opened 2015-10-06
Don't tokenize as 'import' after ';'

### LINKED ISSUE #235: [ES6] Highlight error when there's a comment after the import
Hopefully this is the right project to report this issue to...

MaximSokolov:
Fixes #235

### DIFF
--- a/grammars/javascript.cson
+++ b/grammars/javascript.cson
@@ -41,10 +41,7 @@
         'name': 'meta.brace.curly.js'
       }
     ]
-    'end': '(;)?$'
-    'endCaptures':
-      '1':
-        'name': 'punctuation.terminator.statement.js'
+    'end': '(?=;|$)'
     'name': 'meta.import.js'
   }
   {
--- a/spec/javascript-spec.coffee
+++ b/spec/javascript-spec.coffee
@@ -652,9 +652,9 @@ describe "Javascript grammar", ->
       {tokens} = grammar.tokenizeLine('import point; // comment')
       expect(tokens[0]).toEqual value: 'import', scopes: ['source.js', 'meta.import.js', 'keyword.control.js']
       expect(tokens[2]).toEqual value: 'point', scopes: ['source.js', 'meta.import.js']
-      expect(tokens[3]).toEqual value: '; ', scopes: ['source.js', 'meta.import.js']
-      expect(tokens[4]).toEqual value: '//', scopes: ['source.js', 'meta.import.js', 'comment.line.double-slash.js', 'punctuation.definition.comment.js']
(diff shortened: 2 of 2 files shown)


grammars/javascript.cson
```diff
@@ -41,7 +41,7 @@
         'name': 'meta.brace.curly.js'
       }
     ]
-    'end': '(;)?$'
+    'end': '(\\;)|$'
```
50Wliu: I don't think this needs to be escaped

MaximSokolov: Fixed

spec/javascript-spec.coffee
```diff
@@ -652,9 +652,9 @@ describe "Javascript grammar", ->
       {tokens} = grammar.tokenizeLine('import point; // comment')
       expect(tokens[0]).toEqual value: 'import', scopes: ['source.js', 'meta.import.js', 'keyword.control.js']
       expect(tokens[2]).toEqual value: 'point', scopes: ['source.js', 'meta.import.js']
-      expect(tokens[3]).toEqual value: '; ', scopes: ['source.js', 'meta.import.js']
-      expect(tokens[4]).toEqual value: '//', scopes: ['source.js', 'meta.import.js', 'comment.line.double-slash.js', 'punctuation.definition.comment.js']
-      expect(tokens[5]).toEqual value: ' comment', scopes: ['source.js', 'meta.import.js', 'comment.line.double-slash.js']
+      expect(tokens[3]).toEqual value: ';', scopes: ['source.js', 'meta.import.js', 'punctuation.terminator.statement.js']
```
50Wliu: This probably shouldn't be tagged as `meta.import.js` anymore.  What do you think?

MaximSokolov: I agree. It isn't correct ES6 code. It can be:
```js
import "point";
import point from "module-name";
import { point } from "module-name";
etc.
```
I am working on ES6 import/export. PR will be soon

MaximSokolov: If you meant `;` I don't think it's possible to not tag it since `;` is the end of begin/end pattern

grammars/javascript.cson
```diff
@@ -41,7 +41,7 @@
         'name': 'meta.brace.curly.js'
       }
     ]
-    'end': '(;)?$'
+    'end': '(;)|$'
```
50Wliu: Related to my comment below (or above I guess, depending on how you look at it): Can this be made into `(?=;|$)` to prevent the `;` from also getting tokenized as `meta.import.js` and also to remove the `endCaptures` block?

MaximSokolov: Seems it works. I'll do it in a separate PR

MaximSokolov: Done

50Wliu:
Thanks!

PR #237, opened 2015-10-06
ES6 import export

### LINKED ISSUE #232: export/import default as
Anyone else find this strange?
```js
import { default as foo } from 'module';
export { default as foo } from 'module';
```
screen shot 2015-10-02 at 15 42 24

Why is `as` highlighted in the import statement, but `default` highlighted in the export?
In this situation, shouldn't `as` be highlighted in both, and `default` in neither? `default` is used as an identifier after all, not as a keyword.

MaximSokolov:
Fixes #232
```js
export var x = 0;
export const CONSTANT = 0;
export let scopedVariable = 0;
export function func(p1, p2){}
export class Foo {}
export { bar };
export { bar, foo as alias };
export default "string";
export default 123 + 23;
export default name;
export default function () {}
export default function func() {}
export default class {}
export default class Foo {}
export { foo as default };
export { name } from 'module-name';
export * from 'module-name';
export { default as alias } from 'module-name';
export { default as alias }; //illegal
export { name as * }; //illegal
import "module-name";
import * as name from "module-name"
import defaultMember from "module-name";
import { member } from "module-name";
import { default as defaultMember } from "module-name";
import { member1 , member2 as alias2 } from "module-name";
import { member1 , member2 as default } from "module-name"; //illegal
import { member1 , member2 as * } from "module-name"; //illegal
import defaultMember, { member } from "module-name";
import defaultMember, * as alias from "module-name";
import defaultMember, * as alias from "module-name";
```
atom-dark
> Atom Dark

one-dark
> One Dark

monokai
> Monokai

### DIFF
--- a/grammars/javascript.cson
+++ b/grammars/javascript.cson
@@ -116,6 +116,137 @@
     'name': 'meta.import.js'
   }
   {
+    'comment': 'ES6 export'
+    'begin': '(?<!\\.)\\b(export)(?!\\s*:)\\b'
+    'beginCaptures':
+      '1':
+        'name': 'keyword.control.js'
+    'patterns': [
+      {
+        'include': '#numbers'
+      }
+      {
+        'comment': '`{ member1 , member2 as alias2 , [...] }` inside re-export'
+        'begin': '\\{(?=.*\\bfrom\\b)'
+        'beginCaptures':
+          0:
+            'name': 'punctuation.definition.modules.begin.js'
+        'end': '\\}'
+        'endCaptures':
+          0:
+            'name': 'punctuation.definition.modules.end.js'
+        'patterns': [
+          {
+            'comment': '(default|name) as alias'
+            'captures':
+              '1':
+                'name': 'variable.language.default.js'
+              '2':
+                'name': 'variable.other.module.js'
+              '3':
+                'name': 'keyword.control.js'
+              '4':
+                'name': 'variable.language.default.js'
+              '5':
+                'name': 'invalid.illegal.js'
+              '6':
+                'name': 'variable.other.module-alias.js'
+            'match': '(?x)
+              (?: \\b(default)\\b | \\b([a-zA-Z_$][a-zA-Z_$0-9]*)\\b)
+              \\s*
+              (\\b as \\b)
+              \\s*
+              (?: \\b(default)\\b | (\\*) | \\b([a-zA-Z_$][a-zA-Z_$0-9]*)\\b)
+            '
+          }
+          {
+            'match': ','
+            'name': 'meta.delimiter.object.comma.js'
+          }
+          {
+            'match': '\\b([a-zA-Z_$][a-zA-Z_$0-9]*)\\b'
+            'name': 'variable.other.module.js'
+          }
+        ]
+      }
+      {
+        'comment': '{ member1 , member2 as alias2 , [...] }'
+        'begin': '(?![a-zA-Z_$0-9])\\{'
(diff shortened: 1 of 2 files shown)


grammars/javascript.cson
```diff
@@ -14,35 +14,237 @@
...
+              (?: \\b(default)\\b | \\b([a-zA-Z_$][a-zA-Z_$0-9]*)\\b)
+              \\s*
+              (\\b as \\b)
+              \\s*
+              (?: (\\b default \\b | \\*) | \\b([a-zA-Z_$][a-zA-Z_$0-9]*)\\b)
+            '
+          }
+          {
+            'match': ',',
+            'name': 'punctuation.separator.modules.js'
```
MaximSokolov: I'll remove these needless commas (after `'match': ','`)

Victorystick:
This looks really good! Just one question; is there any reason that the `default` in Atom Dark and One Dark are highlighted differently in

```js
export { foo as default };
export { default as alias } from 'module-name';
```

but the same (red) in Monokai?

MaximSokolov:
`default`, `*` are `variable.language` (variables containing default module/all modules):
```js
export * from 'module-name';
import * as name from "module-name"
export { default as defaultMember } from 'module-name';
import { default as defaultMember } from "module-name";
```
`default` is `storage.modifier `:
```js
export default foo;
export { foo as default };
```
`default` "modifies" `foo` i.e. specifies that `foo` is default module

Victorystick:
Actually, there was a discussion in an ES forum whether to add an equals sign after `export default` to make it more obvious that is a special, non-local 'default' variable that is declared using that syntax: `export default = foo`. So `default` isn't really a modifier in this case.

In the second case, you export the binding `foo` from another module under the name `default`. There is no "modification" going on, just another binding to the same variable.

I therefore think that `default` should have the scope `variable.language` in the latter two examples as well.

grammars/javascript.cson
```diff
@@ -14,35 +14,237 @@
...
+        'match': '\\b(from)\\b',
+        'name': 'keyword.control.js'
+      }
+      {
+        'match': '\\b([a-zA-Z_$][a-zA-Z_$0-9]*)\\b(?=.*\\bfrom\\b)'
+        'name': 'variable.other.module.js'
+      }
+      {
+        'match': ',',
+        'name': 'punctuation.separator.modules.js'
```
50Wliu: Why is this not `meta.delimiter.object.comma.js`?

grammars/javascript.cson
```diff
@@ -633,6 +833,14 @@
   }
 ]
 'repository':
+  'operators':
```
50Wliu: If we're going to make this a repository it should probably go beneath `numbers`

MaximSokolov:
Updated:
atom-dark-2
one-dark-2

grammars/javascript.cson
MaximSokolov: I did it to include operators in 'export' pattern:
```js
export default 123 + 23;
export var x = 0;
etc.
```
Is order important in 'repository' section?

thomasjo:
Love it! Unless anyone else has anything to add, this looks ready to :ship:

Victorystick:
I'm sold. :shipit:

grammars/javascript.cson
50Wliu: Not exactly, but it makes it look nicer :).

50Wliu:
:tada:

PR #248, opened 2015-11-05
Capture regular braces before ending interpolation

### LINKED ISSUE #199: template string bracket mismatch
Hey atom peeps. In the solarized dark repo we have a cool pull request that will be adding cool template string coloring for javascript <a href="https://github.com/atom/solarized-dark-syntax/pull/48">see here</a> and there was a bug uncovered showing that template strings do not match up closing brackets properly.

For ex:

in our case, the open bracket color is dark blue (which is correct) and the closing bracket is cyan, the closing bracket should be red and the bracket closing the function should be blue and not cyan. Would love to help out with this but I am a tad unfamiliar with how the language packages actually work. If anyone could help on this let me know! Thanks

50Wliu:
Fixes #199.  Also includes some very minor style fixes.

In the future I'd like to combine all the `meta.brace.*.js` patterns into their own `braces` block, but that was too much work for this.

/cc @schmittyjd, @methodgrab

### DIFF
--- a/grammars/javascript.cson
+++ b/grammars/javascript.cson
@@ -670,7 +670,7 @@
         'name': 'meta.brace.round.js'
       }
       {
-        'match': '\\{|\\}'
+        'match': '{|}'
         'name': 'meta.brace.curly.js'
       }
       {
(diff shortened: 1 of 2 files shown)


kevinsawicki:
:+1:

PR #249, opened 2015-11-05
Don't capture whitespace as `storage.type.arrow.js`

50Wliu:
Self-explanatory.  Only the arrow itself should be captured.


50Wliu:
Note: if this is merged before #248 then the spec for that will have to be updated, and vice versa.

kevinsawicki:
I've merged #248, does this PR need to be updated?

50Wliu:
Yep, updated!

kevinsawicki:
:+1: :ship:

PR #250, opened 2015-11-05
Don't tokenize numbers that follow a $ or _

### LINKED ISSUE #234: Numbers after $ in identifiers names breaks syntax
```js
var endsWith$1;
```
screen shot 2015-10-05 at 15 43 33

50Wliu:
Obviously a better fix would be to properly highlight identifiers in the first place, but as previously noted that is a much bigger effort.

Fixes #234


kevinsawicki:
:+1:

PR #254, opened 2015-11-09
Refactoring

MaximSokolov:


MaximSokolov:
I'll divide it into several PRs so it would be easy to review

PR #257, opened 2015-11-17
improve function snippet

Hurtak:
- `function_name` renamed to `functionName` because calemCase is more common in JavaScript world
- removed `argument` while keeping tab stop 
  - improves writing of functions with no parameters
  - does not change writing of functions with 1+ parameters
- removed comment `// body ...` because it's useless and sometimes it slows us down when we are not using tab stops, if we are using tab stops experience is the same

comparison of 2 snippets


50Wliu:
Can you fix the other snippets as well that are using underscore_case as well as `//body`?

Hurtak:
fixed

50Wliu:
Looks like there's some conflicts - can you please rebase?  Thanks!

Hurtak:
rebased

50Wliu:
:+1:

PR #260, opened 2015-11-18
Revert argumetns matching

MaximSokolov:
To properly tokenize arguments, functions' declarations (anything else?) should be tokenized first
Atom specs should be reverted as well...

Ref: #224 (Recognize function bodies)

### DIFF
--- a/grammars/javascript.cson
+++ b/grammars/javascript.cson
@@ -1056,49 +1056,7 @@
   'arguments':
     'patterns': [
       {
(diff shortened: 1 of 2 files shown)


MaximSokolov:
I checked atom/atom#9655 It seems it shouldn't affect atom core specs, so new version can be released

50Wliu:
I'll be publishing a new version once I get home.

PR #265, opened 2015-11-23
Fix not a constant properties

MaximSokolov:


50Wliu:
What exactly does this fix?

MaximSokolov:
`P` in `obj.Property` is tokenized as `constant.other.property`

50Wliu:
Oh, well that's embarrassing.

PR #272, opened 2015-11-27
Allow $dollar$ signs in identifiers

### LINKED ISSUE #268: $exports isn't highlighted correctly
An extract from [zloirock/core-js](https://github.com/zloirock/core-js) that breaks highlighting.

screen shot 2015-11-26 at 11 29 07

```js
}
$export($export.S + $export.F * !DESCRIPTORS, 'Object', {
  // 19.1.2.6 / 15.2.3.3 Object.getOwnPropertyDescriptor(O, P)
  getOwnPropertyDescriptor: $.getDesc,
  // 19.1.2.4 / 15.2.3.6 Object.defineProperty(O, P, Attributes)
  defineProperty: $.setDesc,
  // 19.1.2.3 / 15.2.3.7 Object.defineProperties(O, Properties)
  defineProperties: defineProperties
});

  // IE 8- don't enum bug keys
var keys1 = ('constructor,hasOwnProperty,isPrototypeOf,propertyIsEnumerable,' +
            'toLocaleString,toString,valueOf').split(',')
  // Additional keys for getOwnPropertyNames
  , keys2 = keys1.concat('length', 'prototype')
```

### LINKED ISSUE #270: Function/method calls are not tokenized correctly when started with $
`$` should be part of `entity.name.function.js`
```js
$functionCall()
obj.$methodCall()
```

MaximSokolov:
```js
0illegal()
$abc$()
$()
$obj$.$abc$()
$.$()
$.$
$.$prop$
$.CON$TANT
$.C$
class $
class $abc$ extends $
class extends $abc
```
> One Dark Syntax

Before:
before
After:
after

Fixes #270 
Fixes #268


grammars/javascript.cson
```diff
@@ -1096,11 +1105,13 @@
     'patterns': [
       {
         'comment': 'functionCall(arg1, "arg2", [...])'
-        'begin': '\\b(\\w+)\\s*(\\()'
+        'begin': '(?:(\\d\\w*)|([a-zA-Z_$][\\w$]*))\\s*(\\()'
         'beginCaptures':
           '1':
-            'name': 'entity.name.function.js'
+            'name': 'invalid.illegal.js'
```
MaximSokolov: Without `invalid.illegal` it would be `...(\\$+|(?:\\$+)?\\b[a-zA-Z_$][\\w$]*)...`

grammars/javascript.cson
```diff
@@ -513,7 +513,16 @@
...
+    'match': '''(?x)
+      \\b(class)
+      (?:
+        (?: \\s+(extends) \\s+ ([a-zA-Z_$][\\w$]*) )
+        |
+        (?: (?: \\s+ ([a-zA-Z_$][\\w$]*) )
+          (?: \\s+ (extends) \\s+ ([a-zA-Z_$][\\w$]*) )?
+        )
+      )
+    '''
```
MaximSokolov: btw, should I rebase #254?

Victorystick:
I'm loving this!

grammars/javascript.cson
50Wliu: We haven't decided whether or not to proceed with that one yet.  If we do, I'll ask you to :).

Also, can you remove the spaces in this regex?  I think they attract too much attention to themselves for some reason.

MaximSokolov: done

50Wliu:
:shaved_ice:

PR #277, opened 2015-12-03
Tokenize let and var as storage.type.var.js

### LINKED ISSUE #201: Follow naming conventions

50Wliu:
#206 is back! :tada:

Tested my changes locally and I *think* I can get everything working this time (it's hard to tell because I get ~50 spec failures on atom/atom).  Will be pushing PRs to language-html and atom shortly.

/cc @atom/languages @slinto

Actually fixes #201

### DIFF
--- a/grammars/javascript.cson
+++ b/grammars/javascript.cson
@@ -646,8 +646,12 @@
     'name': 'storage.type.js'
   }
   {
-    'match': '(?<!\\.)\\b(async|export|extends|implements|let|private|protected|public|static|var)(?!\\s*:)\\b'
+    'match': '(?<!\\.)\\b(async|export|extends|implements|private|protected|public|static)(?!\\s*:)\\b'
     'name': 'storage.modifier.js'
+  }
+  {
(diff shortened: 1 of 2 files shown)


jerone:
Shouldn't `const` be included in that list of variable declaration types too? 

Ref: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements#Declarations
Ref: http://www.ecma-international.org/ecma-262/6.0/#sec-let-and-const-declarations

50Wliu:
Maybe not `storage.type.var`...does `storage.type.constant` sound better? /cc @MaximSokolov

MaximSokolov:
`storage.type.const` or just `storage.type` as styles for `constant` can affect "constant" keyword 
(see [atom/language-go#63 (comment)](https://github.com/atom/language-go/pull/63#issuecomment-141048632))

Related question: Is `const` in `const int i;` a "modifier" or it's `storage.type` as well?

PR #278, opened 2015-12-05
Add exports & case snippet.

luop90:
...this wasn't supposed to be both the exports and the case snippets. But it appears GitHub added both anyway.

But basically, adds the 'case' and the 'exports' snippet.

### DIFF
--- a/snippets/language-javascript.cson
+++ b/snippets/language-javascript.cson
@@ -104,3 +104,6 @@
   'Class':
     'prefix': 'class'
     'body': 'class ${1:ClassName} {\n\tconstructor($2) {\n\t\t$3\n\t}\n}'
+  'Exports Function':
+    'prefix': 'exp'
+    'body': 'exports.${1:functionName} = function ($2) {\n\t${3:// body...}\n};'
--- a/snippets/language-javascript.cson
+++ b/snippets/language-javascript.cson
@@ -89,6 +89,9 @@
   'switch':
     'prefix': 'switch'
     'body': 'switch (${1:expression}) {\n\tcase ${2:expression}:\n\t\t$4\n\t\tbreak;$5\n\tdefault:\n\t\t$3\n}'
+  'case':
+    'prefix': 'case'
+    'body': 'case ${1:expression}: \n\t$2\n\tbreak;'
   'try':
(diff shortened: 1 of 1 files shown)


snippets/language-javascript.cson
```diff
@@ -104,3 +104,9 @@
   'Class':
     'prefix': 'class'
     'body': 'class ${1:ClassName} {\n\tconstructor($2) {\n\t\t$3\n\t}\n}'
+  'Exports Function':
+    'prefix': 'exp'
+    'body': 'exports.${1:functionName} = function ($2) {\n\t${3:// body...}\n};'
+  'Case':
```
50Wliu: I would move this to right below the `switch` snippet and lowercase it as well.

snippets/language-javascript.cson
```diff
@@ -104,3 +104,9 @@
   'Class':
     'prefix': 'class'
     'body': 'class ${1:ClassName} {\n\tconstructor($2) {\n\t\t$3\n\t}\n}'
+  'Exports Function':
```
50Wliu: If we're adding an Exports snippet it might be worthwhile to add an `import` one as well.

luop90:
I've updated with the changes. case is now below switch, all lowercase. Also, there's two different types of exports, so I've made the difference between them in both snippets.

However, with 'import', I have no use experience with it (mainly because it's locked behind the --harmony flag), so I can't accurately provide a snippet that would be useful.

snippets/language-javascript.cson
```diff
@@ -89,6 +89,9 @@
   'switch':
     'prefix': 'switch'
     'body': 'switch (${1:expression}) {\n\tcase ${2:expression}:\n\t\t$4\n\t\tbreak;$5\n\tdefault:\n\t\t$3\n}'
+  'case':
+    'prefix': 'case'
+    'body': 'case ${1:expression}: \n\t$2\n\tbreak;'
```
50Wliu: Unnecessary space at the end of the colon in `case expression: `.

luop90:
Fixed the unneeded space. (I didn't notice that, thanks)

50Wliu:
Thanks for this!

PR #276, opened 2015-12-02
[art] Reorder groups

MaximSokolov:
1. comment
2. match
3. captures
4. name

1. comment
2. begin
3. beginCaptures
4. end
4. endCaptures
5. name
6. patterns

How about:

1. comment
2. __name__
3. match
4. captures

1. comment
1. __name__
2. begin
3. beginCaptures
4. end
4. endCaptures
6. patterns

PR #284, opened 2015-12-15
Added delete token

C-Bouthoorn:
Added delete token as noted [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/delete)

### DIFF
--- a/grammars/javascript.cson
+++ b/grammars/javascript.cson
@@ -268,7 +268,7 @@
         'include': '#operators'
       }
     ]
-    'end': '(?=;|\\bfunction\\b|\\bclass\\b|\\blet\\b|\\bvar\\b|\\bconst\\b|$)'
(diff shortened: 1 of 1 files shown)


50Wliu:
Are you sure this should be under `meta.export.js`?

MaximSokolov:
This rule is supposed to match `var`, `let`... in named exports:
```js
export var x = 0;
export const CONSTANT = 0;
export let scopedVariable = 0;
export function func(p1, p2){}
export class Foo {}
```
btw, I moved these lines up in #276

50Wliu:
Ok, thanks.  I didn't notice that this rule as a whole was for exports.

MaximSokolov:
Sorry, I wasn't clear. `export delete` is not legit

PR #291, opened 2015-12-21
Tokenize spread operator

### LINKED ISSUE #290: Issue with Spread operator
image

Text placed right after dot(s) may not a property

MaximSokolov:
Fixes #290

### DIFF
--- a/grammars/javascript.cson
+++ b/grammars/javascript.cson
@@ -694,7 +694,7 @@
       }
       {
         'match': '\\.\\.\\.'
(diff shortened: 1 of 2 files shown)


50Wliu:
Question...how does this fix #290?  Wouldn't the variable behind the spread operator still be tokenized the same?

MaximSokolov:
See 75eb803

50Wliu:
Ok, thanks!

PR #293, opened 2015-12-27
Add 'Set' and 'Map' to 'support.class.js'

KSXGitHub:


mnquintana:
Would you mind adding specs for this? Thanks!

50Wliu:
@mnquintana I don't think specs are actually needed for this because it's a bit unfeasible to test the entire list (and makes updates/removals a pain).  There's already various tests that match for `support.class.js`, so I think we're fine here :smile:.

PR #295, opened 2015-12-31
Fix function/method calls being superseded

### LINKED ISSUE #294: Function-call highlights being superceded
Currently the highlighting for `meta.method-call > .entity.function` is broken for some named functions matching `support.function.dom`.

```js
a.fn(); // "fn" can be styled as a function being called
a.submit() // "submit" is only given a `support.function.dom` scope, no wrapped `.method-call`
a.Math.abs() // Same as above
```
[The offending rules](https://github.com/atom/language-javascript/blob/master/grammars/javascript.cson#L794).

These scopes should ideally be supplementary, though I realize it's not easy.
I'll try re-ordering the rules and fiddle around but I'm not sure how to fix this just yet.

nfour:
Fixed by putting function constants into their own repository, then including it within `method_call` and `function_call` repos.

This produces the below markup:
```html
<span class="meta method-call js">
  <span class="meta delimiter method period js">.</span>
  <span class="entity name function js">
    <span class="support function dom js">getElementById</span>
  </span>
  <span class="punctuation definition arguments begin js">(</span>
  <span class="punctuation definition arguments end js">)</span>
</span>
```

Fixes #294


nfour:
It's currently failing due to the nested nature produced, always wrapping in `entity.name.function.js`.

It's an easy enough fix to have it replace the scope entirely, just not sure which way is best.

MaximSokolov:
Could you also add tests for built-in methods? Something like:
```coffee
describe "method calls", ->
  ...
  describe "built-in methods", ->
    methods = ['shift', 'showModelessDialog', ...]
    domMethods = ['substringData', 'submit', ...]

    for method in methods
      it "tokenizes '#{method}'", ->
        {tokens} = grammar.tokenizeLine('.' + method + '()')
        expect(tokens[0]).toEqual value: '.', scopes: ['source.js', 'meta.method-call.js', 'meta.delimiter.method.period.js']
        expect(tokens[1]).toEqual value: method, scopes: ['source.js', 'meta.method-call.js', 'support.function.js']
        expect(tokens[2]).toEqual value: '(', scopes: ['source.js', 'meta.method-call.js', 'punctuation.definition.arguments.begin.js']
        expect(tokens[3]).toEqual value: ')', scopes: ['source.js', 'meta.method-call.js', 'punctuation.definition.arguments.end.js']

    for domMethod in domMethods
      it "tokenizes '#{domMethod}'", ->
        {tokens} = grammar.tokenizeLine('.' + domMethod + '()')
        expect(tokens[0]).toEqual value: '.', scopes: ['source.js', 'meta.method-call.js', 'meta.delimiter.method.period.js']
        expect(tokens[1]).toEqual value: domMethod, scopes: ['source.js', 'meta.method-call.js', 'support.function.dom.js']
        expect(tokens[2]).toEqual value: '(', scopes: ['source.js', 'meta.method-call.js', 'punctuation.definition.arguments.begin.js']
        expect(tokens[3]).toEqual value: ')', scopes: ['source.js', 'meta.method-call.js', 'punctuation.definition.arguments.end.js']
```

nfour:
Thanks that seems like a good place to start; will do.

nfour:
Alright seems good now!

grammars/javascript.cson
```diff
@@ -1096,7 +1080,11 @@
           '1':
             'name': 'meta.delimiter.method.period.js'
           '2':
-            'name': 'entity.name.function.js'
+            'patterns': [
```
50Wliu: I think(?) that this can just be `'include': '#functions'`, without the extra `patterns` block.

spec/javascript-spec.coffee
```diff
@@ -1104,6 +1104,26 @@ describe "Javascript grammar", ->
...
...", "eval", "elementFromPoint", "exp", "exec", "execScript", "execCommand", "valueOf", "UTC", "queryCommandState", "queryCommandIndeterm", "queryCommandEnabled", "queryCommandValue", "find", "file", "fileModifiedDate", "fileSize", "fileCreatedDate", "fileUpdatedDate", "fixed", "fontsize", "fontcolor", "forward", "floor", "fromCharCode", "watch", "link", "load", "log", "lastIndexOf", "asin", "anchor", "acos", "attachEvent", "atob", "atan", "atan2", "apply", "alert", "abs", "abort", "round", "routeEvents", "resize", "resizeBy", "resizeTo", "recalc", "returnValue", "replace", "reverse", "reload", "releaseCapture", "releaseEvents", "random", "go", "get", "getMilliseconds", "getSeconds", "getMinutes", "getHours", "getMonth", "getDay", "getYear", "getFullYear", "getTime", "getDate", "getTimezoneOffset", "getUTC", "getUTCMilliseconds", "getUTCSeconds", "getUTCMinutes", "getUTCHours", "getUTCDay", "getUTCMonth", "getUTCFullYear", "getUTCDate", "getAttention", "getSelection", "getResponseHeader", "getAllResponseHeaders", "min", "moveBy", "moveBelow", "moveTo", "moveToAbsolute", "moveAbove", "mergeAttributes", "match", "margins", "max", "btoa", "big", "bold", "borderWidths", "blink", "back" ]
```
50Wliu: No need to have all of these, just test one or two.  Same with `domMethods`.

grammars/javascript.cson
```diff
@@ -1135,6 +1127,26 @@
...
+            'name': 'support.function.js'
+      }
+      {
+        'match': '\\b(substringData|submit|splitText|setNamedItem|setAttribute|setAttributeNode|select|hasChildNodes|hasFeature|namedItem|click|close|cloneNode|createComment|createCDATASection|createCaption|createTHead|createTextNode|createTFoot|createDocumentFragment|createProcessingInstruction|createEntityReference|createElement|createAttribute|tabIndex|insertRow|insertBefore|insertCell|insertData|item|open|deleteRow|deleteCell|deleteCaption|deleteTHead|deleteTFoot|deleteData|focus|write|writeln|add|appendChild|appendData|reset|replaceChild|replaceData|move|moveNamedItem|moveChild|moveAttribute|moveAttributeNode|getNamedItem|getElementsByName|getElementsByTagName|getElementById|getAttribute|getAttributeNode|blur)\\b'
+        'captures':
+          '1':
+            'name': 'support.function.dom.js'
+      }
+      {
+        'match': ".*"
```
50Wliu: This should probably be `.+` so we don't have any empty matches.

grammars/javascript.cson
nfour: Tried this, it does actually need the patterns property.

grammars/javascript.cson
nfour: True. I also refactored the `captures` out as `\b` wont be captured.

nfour:
I was also looking at these references, I think they could be merged into the `#functions` repo as well.

- [support.function.event-handler.js](https://github.com/nfour/language-javascript/blob/patch-1/grammars/javascript.cson#L807)
- `require` [support.function.js](https://github.com/nfour/language-javascript/blob/patch-1/grammars/javascript.cson#L8110)
- [support.function.js.console](https://github.com/nfour/language-javascript/blob/patch-1/grammars/javascript.cson#L623) (and shouldn't it be `support.function.console.js`?)

grammars/javascript.cson
50Wliu: Unfortunately there's no docs.  Most of it is like you said directly from Textmate, and the majority of Atom's builtin syntaxes were converted from the Textmate ones.

nfour:
It appears I also need to refactor the `support.function.js` regex list. It contains method-calls and function-calls together, but `parseInt` or `setInterval` for example shouldn't be recognized as a support function in this context: `this.parseInt()`.

grammars/javascript.cson
```diff
@@ -1135,6 +1134,36 @@
...
+      {
+        'match': '\\b(substringData|submit|splitText|setNamedItem|setAttribute|setAttributeNode|select|hasChildNodes|hasFeature|namedItem|click|close|cloneNode|createComment|createCDATASection|createCaption|createTHead|createTextNode|createTFoot|createDocumentFragment|createProcessingInstruction|createEntityReference|createElement|createAttribute|tabIndex|insertRow|insertBefore|insertCell|insertData|item|open|deleteRow|deleteCell|deleteCaption|deleteTHead|deleteTFoot|deleteData|focus|write|writeln|add|appendChild|appendData|reset|replaceChild|replaceData|move|moveNamedItem|moveChild|moveAttribute|moveAttributeNode|getNamedItem|getElementsByName|getElementsByTagName|getElementById|getAttribute|getAttributeNode|blur)\\b'
+        'name': 'support.function.dom.js'
+      }
+      {
+        'match': ".+"
+        'name': 'entity.name.function.js'
+      }
+    ]
+  'function_calls':
```
50Wliu: Can we rename this `functions`?  I find it confusing to have both `function_call` and `function_calls`.

grammars/javascript.cson
```diff
@@ -1135,6 +1134,36 @@
         ]
       }
     ]
+  'function_methods':
```
50Wliu: Would it be better to rename this `method_functions` (or maybe even `methods`), and the one below it just `functions`?  I think that clears up some of the ambiguity between `function_call` and `function_calls`.

PR #296, opened 2015-12-31
Revert "Added delete token"

MaximSokolov:
Reverts atom/language-javascript#284


50Wliu:
You know, that's what I thought...but then the docs aren't always that clear, and since none of the major browsers support export/import yet, I couldn't check either... :(

PR #301, opened 2016-01-11
Updated regex's for indentation

chfritz:
This is intended for use with the new language-mode logic of atom/atom
commit atom/atom@544112516e362f8741f418a9876b36d9261dd13b.
(See PR https://github.com/atom/atom/pull/10384).

Fixes atom/atom#4983 (when used with the above mentioned commit in atom/atom).


chfritz:
https://github.com/atom/atom/pull/10384, which is required for this, was closed. The new plan is https://github.com/atom/language-javascript/issues/594.

PR #307, opened 2016-01-18
[WIP] Add support for object keys

MaximSokolov:
This **wouldn't** work with:
```js
// `prop` are not tagged
? {prop: value} : c

// labels would be tagged as object keys
loop1:
for (i = 0; i < 3; i++) {      //The first for statement is labeled "loop1"
   loop2:
   for (j = 0; j < 3; j++) {   //The second for statement is labeled "loop2"
      if (i === 1 && j === 1) {
         continue loop1;
      }
      console.log("i = " + i + ", j = " + j);
   }
}
```
Tested with:
```js
prop:
"prop":
null:
ABC:
debugger:
module:
super:
this:
class:
NaN:
{a: b, c: "d", "e": f}

? (a.(123)) : (b("c"))
? "a" : b
? a : "b"
? module : null
? this : "test"

case "a":
case a:
case a + b:
case a.b:
case 123:
case foo(bar):
case NaN:
case Area:
case 123:
case this:
default:
```
What do you think about the approach? It looks really hacky


50Wliu:
I mean, this does seem to clear up some of the regexes.  The case/default stuff is definitely something that can be improved though, and may also allow us to match labels correctly.  Can you try putting that in a separate repository (maybe `'include': '#switch-case'`), and only call that when we hit a `switch` statement?

MaximSokolov:
Objects literals inside ternary expression fixed by using `begin`/`end` matching for curly brackets. Looking for following word in key pattern allows to match labels properly. Switch statement moved to a repo.

50Wliu:
@MaximSokolov So is this ready for review, or are there more edge cases that need to be addressed?

MaximSokolov:
It's ready for review. However PR brings highlighting changes, so I want to add before/after info and steps to change colors before merging

grammars/javascript.cson
```diff
@@ -17,8 +17,162 @@
...
 'patterns': [
   {
+    'include': '#switch_statement'
+  }
+  {
+    # foo: function() { … }
+    'begin': '''(?x)
+      (\\b([a-zA-Z_?.$][\\w?.$]*))
+      \\s*(:)\\s*
+      \\b(?:(async)(?:\\s+))?
```
50Wliu: Can't this be `(\\basync\\b)?` instead?

grammars/javascript.cson
```diff
@@ -17,8 +17,162 @@
...
+      (function\\*?)
+      \\s*(\\*?)\\s*
```
50Wliu: Wouldn't this combined with the previous line give us `function**`?  Only one asterisk is allowed.

grammars/javascript.cson
```diff
@@ -17,8 +17,162 @@
...
+      ([a-zA-Z_?$][\\w?$]*)?
```
50Wliu: I don't think this line is correct either, as object-literal functions have to be anonymous.

grammars/javascript.cson
```diff
@@ -17,8 +17,162 @@
...
+  }
+  {
+    # "foo": function() { … }
+    'begin': '''(?x)
+      ((?:((\')([^\']*?)(\'))|((")([^"]*?)("))))
+      \\s*(:)\\s*
+      \\b(?:(async)(?:\\s+))?
+      (function\\*?)
+      \\s*(\\*?)\\s*
+      ([a-zA-Z_?$][\\w?$]*)?
```
50Wliu: Same comments as above

grammars/javascript.cson
```diff
@@ -17,8 +17,162 @@
...
+        'name': 'storage.modifier.js'
+      '5':
+        'name': 'storage.type.function.js'
+      '6':
+        'name': 'storage.type.function.js'
+      '7':
+        'name': 'entity.name.function.js'
+      '8':
+        'name': 'punctuation.definition.parameters.begin.js'
+    'end': '(\\))'
```
50Wliu: Why not just `\\)` and match the 0 capture?

grammars/javascript.cson
```diff
@@ -17,8 +17,162 @@
...
+      {
+        'match': '(\\w+)(?=\\s*:)'
+        'captures':
+          '1':
+            'patterns': [
+              'include': '$self'
+            ]
+      }
+      {
+        'match': '(?:"[^"]*")|(?:\'[^\']*\')(?=\\s*:\\s*\\.*)'
```
50Wliu: Why do we need this?

grammars/javascript.cson
```diff
@@ -17,8 +17,162 @@
...
+              'include': '#strings'
+            ]
+      }
+      {
+        'include': '$self'
+      }
+    ]
+  }
+  {
+    'match': '(?:^\\s*)([a-zA-Z_$][\\w$]*)\\s*(:)(?=\\n)'
```
50Wliu: Will this capture succeed if there's a comment after the label, eg `label: /* comment! */`?

grammars/javascript.cson
```diff
@@ -1251,3 +1322,82 @@
         ]
       }
     ]
+  'switch_statement':
+    'patterns': [
+      {
+        'begin': '\\b(switch)\\b'
```
50Wliu: No need for parentheses

grammars/javascript.cson
```diff
@@ -1251,3 +1322,82 @@
...
+            'begin': '{'
+            'beginCaptures':
+              '0':
+                'name': 'punctuation.definition.section.switch-block.begin.js'
+            'end': '(?=})'
+            'patterns': [
+              {
+                'begin': '\\b(case)\\b'
+                'end': ':'
+                'beginCaptures':
```
50Wliu: `beginCaptures` should go before `end`

grammars/javascript.cson
```diff
@@ -1251,3 +1322,82 @@
...
+          }
+          {
+            'begin': '{'
+            'beginCaptures':
+              '0':
+                'name': 'punctuation.definition.section.switch-block.begin.js'
+            'end': '(?=})'
+            'patterns': [
+              {
+                'begin': '\\b(case)\\b'
```
50Wliu: No need for parentheses

grammars/javascript.cson
```diff
@@ -1251,3 +1322,82 @@
...
+                'end': ':'
+                'beginCaptures':
+                  '0':
+                    'name': 'keyword.control.case.js'
+                'endCaptures':
+                  '0':
+                    'name': 'punctuation.definition.section.case-statements.js'
```
50Wliu: Should this be `case-statement` instead?

grammars/javascript.cson
```diff
@@ -1251,3 +1322,82 @@
...
+                  {
+                    'match': '(\\w+)(?=\\s*:)'
+                    'captures':
+                      '1':
+                        'patterns': [
+                          'include': '$self'
+                        ]
+                  }
+                  {
+                    'match': '(?:".*")|(?:\'.*\')(?=\\s*:\\s*\\.*)'
```
50Wliu: Again, why is this needed?

grammars/javascript.cson
```diff
@@ -1251,3 +1322,82 @@
...
+                  }
+                ]
+              }
+              {
+                'match': '(?:^\\s*)?\\b(default)\\b\\s*(:)'
+                'captures':
+                  '1':
+                    'name': 'keyword.control.default.js'
+                  '2':
+                    'name': 'punctuation.definition.section.case-statements.js'
```
50Wliu: Same, I think this should probably be either `case-statement` or `default-statement`

spec/javascript-spec.coffee
```diff
@@ -55,36 +55,12 @@ describe "Javascript grammar", ->
...
-          {tokens} = grammar.tokenizeLine("a ? #{keyword} : b")
-          expect(tokens[2]).toEqual value: ' ', scopes: ['source.js']
-          expect(tokens[3]).toEqual value: keyword, scopes: ['source.js', scope]
-
-        it "tokenizes `#{keyword}` at the end of ternary expressions", ->
-          {tokens} = grammar.tokenizeLine("a ? b : #{keyword}")
-          expect(tokens[4]).toEqual value: ' ', scopes: ['source.js']
-          expect(tokens[5]).toEqual value: keyword, scopes: ['source.js', scope]
+    for keyword in keywords
+      it "tokenize `#{keyword}`", ->
```
50Wliu: tokenizes

spec/javascript-spec.coffee
```diff
@@ -797,23 +736,152 @@ describe "Javascript grammar", ->
...
       expect(tokens[3]).toEqual value: 'yield', scopes: ['source.js', 'meta.control.yield.js', 'keyword.control.js']
 
-    it "tokenizes yield at the end of ternary expressions", ->
-      {tokens} = grammar.tokenizeLine('a ? b : yield')
-      expect(tokens[5]).toEqual value: 'yield', scopes: ['source.js', 'meta.control.yield.js', 'keyword.control.js']
+    it "tokenizes regular expressions inside ternary expressions", ->
+      {tokens} = grammar.tokenizeLine('a ? /b/ : /c/')
+      expect(tokens[0]).toEqual value: 'a ', scopes: ['source.js']
+      expect(tokens[1]).toEqual value: '?', scopes: ['source.js', 'keyword.operator.ternary.js']
+      expect(tokens[2]).toEqual value: ' ', scopes: ['source.js', 'string.regexp.js']
```
50Wliu: This shouldn't be marked as `string.regexp.js` - if it is, that's probably a bug.

spec/javascript-spec.coffee
```diff
@@ -797,23 +736,152 @@ describe "Javascript grammar", ->
...
+      expect(tokens[3]).toEqual value: '/', scopes: ['source.js', 'string.regexp.js', 'punctuation.definition.string.begin.js']
+      expect(tokens[4]).toEqual value: 'b', scopes: ['source.js', 'string.regexp.js']
+      expect(tokens[5]).toEqual value: '/', scopes: ['source.js', 'string.regexp.js', 'punctuation.definition.string.end.js']
+      expect(tokens[6]).toEqual value: ' ', scopes: ['source.js']
+      expect(tokens[7]).toEqual value: ':', scopes: ['source.js', 'keyword.operator.ternary.js']
+      expect(tokens[8]).toEqual value: ' ', scopes: ['source.js', 'string.regexp.js']
```
50Wliu: Ditto

spec/javascript-spec.coffee
```diff
@@ -797,23 +736,152 @@ describe "Javascript grammar", ->
...
+
+      {tokens} = grammar.tokenizeLine('$abc$: 1')
+      expect(tokens[0]).toEqual value: '$abc$', scopes: ['source.js', 'meta.object-key.js', 'variable.other.property.assignment.js']
+      expect(tokens[1]).toEqual value: ':', scopes: ['source.js', 'keyword.operator.assignment.js']
+
+      {tokens} = grammar.tokenizeLine('0abc: 1')
+      expect(tokens[0]).toEqual value: '0abc', scopes: ['source.js', 'meta.object-key.js', 'variable.other.property.assignment.js']
+      expect(tokens[1]).toEqual value: ':', scopes: ['source.js', 'keyword.operator.assignment.js']
+
+      {tokens} = grammar.tokenizeLine('"key": 1')
```
50Wliu: I would also test single quotes and silly keys like "!" or ""

spec/javascript-spec.coffee
```diff
@@ -797,23 +736,152 @@ describe "Javascript grammar", ->
...
+    keywords = ['yield', 'super', 'this', 'null', 'true', 'false', 'debugger', 'exports', '__filename']
+
+    for keyword in keywords
+      it "tokenizes `#{keyword}` when it is an object key", ->
+        {tokens} = grammar.tokenizeLine("#{keyword}: 1")
+        expect(tokens[0]).toEqual value: keyword, scopes: ['source.js', 'meta.object-key.js', 'variable.other.property.assignment.js']
+        expect(tokens[1]).toEqual value: ':', scopes: ['source.js', 'keyword.operator.assignment.js']
+
+  describe "switch statements", ->
+    it "tokenizes `switch` keyword", ->
```
50Wliu: tokenizes the `switch` keyword

spec/javascript-spec.coffee
```diff
@@ -797,23 +736,152 @@ describe "Javascript grammar", ->
...
+      {tokens} = grammar.tokenizeLine('switch(){}')
+      expect(tokens[0]).toEqual value: 'switch', scopes: ['source.js', 'keyword.control.switch.js']
 
-  describe "default: in a switch statement", ->
-    it "tokenizes it as a keyword", ->
-      {tokens} = grammar.tokenizeLine('default: ')
-      expect(tokens[0]).toEqual value: 'default', scopes: ['source.js', 'keyword.control.js']
+    it "tokenizes switch expression", ->
```
50Wliu: expressions

spec/javascript-spec.coffee
```diff
@@ -797,23 +736,152 @@ describe "Javascript grammar", ->
...
+      {tokens} = grammar.tokenizeLine('switch(1+1){}')
```
50Wliu: Pretty useless switch block :laughing:

grammars/javascript.cson
50Wliu: Also, what do you say about assigning the entire switch statement a scope like `meta.switch-statement.js`, and the body of it `meta.switch-statement.body.js` (same thing with cases)?  This brings us in line with other languages like C and Java.  As an added benefit it'll help autocompletion.

50Wliu:
This will also need to be rebased.

doughsay:
I'd like to mention a special case, just to make sure it gets addressed with these changes when they're done: https://github.com/atom/language-javascript/issues/334#issuecomment-219531841

spec/javascript-spec.coffee
```diff
@@ -144,21 +120,6 @@ describe "Javascript grammar", ->
       expect(tokens[6]).toEqual value: '/', scopes: ['source.js', 'string.regexp.js', 'punctuation.definition.string.end.js']
       expect(tokens[7]).toEqual value: ']', scopes: ['source.js', 'meta.brace.square.js']
 
-    it "tokenizes regular expressions inside ternary expressions", ->
-      {tokens} = grammar.tokenizeLine('a ? /b/ : /c/')
-      expect(tokens[ 0]).toEqual value: 'a ', scopes: ['source.js']
-      expect(tokens[ 1]).toEqual value: '?', scopes: ['source.js', 'keyword.operator.js']
```
Rabbss: https://github.com/atom/language-javascript/pull/307/files/6dce6b145dae26139a7f31906ffd23921b6cea50#diff-3424dea2bf931ab4d3fd094e48ad0f35

PR #310, opened 2016-01-21
Add auto detection for Google Apps Script

Spencer-Easton:
Google Apps Script is a Javascript based scripting language for the Google Apps platform. Gs is the file type used.


grammars/javascript.cson
```diff
@@ -1,5 +1,6 @@
 'scopeName': 'source.js'
 'fileTypes': [
+  'gs'
```
50Wliu: This should be moved under `es6` so the list remains alphabetical.

Issue #197, opened 2015-08-12
spread operator regex bug


bucaran:
I think `arr` should be just black.

```js
foo(...arr)
```

50Wliu:
@bucaran It's showing up as plain-old `source.js` for me...are you still seeing something different?

50Wliu:
No response, closing.

bucaran:
I missed this notification. Thanks, it's showing correctly now.

PR #312, opened 2016-01-25
Improve functions

### LINKED ISSUE #193: Arrow functions with a single param as default arguments to functions
```js
// incorrect highlighting - arrow isn't highlighted
function f (a = _ => _) { }
function f (a = arg => arg) { }
function f (a = ret => ret) { }
function f (a = pchaigno => pchaigno) { }

// correct highlighting - arrow is highlighted
function f (a = (_) => _) { }
function f (a = (arg) => arg) { }
function f (a = (ret) => ret) { }
function f (a = (pchaigno) => pchaigno) { }
```

### LINKED ISSUE #223: Function call recognized as method definition
This function call is recognized as a method definition:
```js
write("){");
```
It even [is in the tests](https://github.com/atom/language-javascript/blob/master/spec/javascript-spec.coffee#L518-L525) since f579460ff (the scopes shouldn't contain `meta.method.js`).
@50Wliu Is it a mistake or was it placed here as a bug to fix?

This is happening since #134 which added support for this kind of method definitions:
```js
var foo = {
  baz() {}
};
```

I'm working on a fix. ~~It will required to recognize JavaScript objects (`{}`) which requires to improve the current support for function and method definitions (to not mistake a function body for an object definition).~~ Actually, that won't fix it ~~and I found an other way :-)~~

EDIT: That probably should be a wontfix: I don't see how we can distinguish a function call from a method definition inside an object when there is `"){"` in argument.

### LINKED ISSUE #303: Wrong highlight for a word contains "function"
```javascript
var create = module.exports = functionizeClass(Delegate);
```

image

"function" here is inside another word and should not be highlighted

### LINKED ISSUE #309: Support class not consistent
Support classes are not being marked consistently:

ss

`Ember` is first being marked as a `support.class`, while later it's marked as a plain `variable.other.object`.

### LINKED ISSUE #322: Wrong scope for function names following language variables
See the following differences:

image

The `register` method on `this` should also be marked `entity.name.function.js`

MaximSokolov:
```js
obj.foo = function(){}
obj.prototype.foo = bar;
obj.prototype = { … }
function* foo(){}
```
__Before__:
`obj` – `support.class`
`*` – `storage.type`
`foo` – `entity.name.function.js`

__After__:
`obj` – `variable.other.object`
`*` – `storage.modifier.generator`
`foo` – `variable.other.property`

[Code for testing](https://gist.github.com/MaximSokolov/a473b39c30db90e18e70)

Fixes #303
Fixes #309
Fixes #193
Fixes #223
Fixes #322
~~Closes~~ #197
Replaces and closes #224


grammars/javascript.cson
```diff
@@ -281,220 +281,128 @@
...
     'patterns': [
       {
-        'include': '#function-params'
+        'include': '#functions'
       }
     ]
   }
   {
-    'begin': '([a-zA-Z_?$][\\w?$]*)\\s*(=)\\s*(?:(async)(?:\\s+))?(function\\*?)\\s*(\\*?)\\s*([a-zA-Z_?$][\\w?$]*)?\\s*(\\()'
+    # foo: function()
```
MaximSokolov: [#307 (comment)](https://github.com/atom/language-javascript/pull/307#discussion_r50492922):
> I don't think this line is correct [function name], as object-literal functions have to be anonymous.

`¯\_(ツ)_/¯`
```js
var obj = {foo: function(){}}
console.log(obj.foo); //=> function(){}
obj = {foo: function abc(){}}
console.log(obj.foo); //=> function abc(){}
```
I rather keep it as is

grammars/javascript.cson
```diff
@@ -281,220 +281,128 @@
...
-    'captures':
-      '1':
-        'name': 'support.class.js'
-      '2':
-        'name': 'support.constant.js'
-      '3':
-        'name': 'entity.name.function.js'
-      '4':
-        'name': 'keyword.operator.assignment.js'
-    'comment': 'match stuff like: Sound.prototype.play = myfunc'
```
MaximSokolov: `myfunc` can be not a function:
```js
var Sound = function() {};
myfunc = 123
Sound.prototype.play = myfunc;
var sound = new Sound();
console.log(sound.play); //=> 123
```
Any reason why we should keep this?

50Wliu:
@MaximSokolov Do you want this to be reviewed alongside #307?

MaximSokolov:
@50Wliu It's not necessary

MaximSokolov:
New approach of functions recognition:
```coffee
{
  'begin': '(?=...)'
  'end': '...'
  'name': '3'
  'patterns': [
    {
      'begin': '\\G'
      'end': '...'
      'name': '1'
      'patterns': [ ... ]
    }
    {
      'include': '#function_body' # 2 
    }
  ]
}
```

```
function foo ( ... ) { ... }
^------------------^ |     | - 1
|                    ^-----^ - 2
^--------------------------^ - 3

( ... ) => { ... }
^--------^ |     | - 1
|          ^-----^ - 2
^----------------^ - 3

x => x + x
^--^       - 1, 3
```
All these regions can be scoped. Not sure what profit of nested function bodies' blocks to autocompletion, syntax highlighting; so I'll keep `meta.function` for 1'st region.

[Code for testing](https://gist.github.com/MaximSokolov/a473b39c30db90e18e70)

:warning: Still WIP
Fixes #193
Fixes #223
Closes #197
Replaces and closes #224

grammars/javascript.cson
```diff
@@ -281,240 +281,256 @@
...
-        'name': 'punctuation.definition.parameters.begin.js'
-    'comment': 'match stuff like: foobar: function() { … }'
-    'end': '(\\))'
-    'endCaptures':
-      '1':
-        'name': 'punctuation.definition.parameters.end.js'
-    'name': 'meta.function.json.js'
+    # [param|(params)] => [expression|{statements}]
+    'begin': '(?=(?<![A-Za-z0-9])((\\(([^\\(\\)]*)?\\))|[\\w$]+)\\s*=>)'
+    'end': '(?!(\\s*{|\\G\\(|\\G[\\w$]+|/\\*))((?=\\s*\\S)|(?<=}))'
```
MaximSokolov: The end pattern definitely should be improved:

the_end

MaximSokolov:
It's ready for review

grammars/javascript.cson
```diff
@@ -281,240 +281,283 @@
...
-        'name': 'punctuation.definition.parameters.begin.js'
-    'comment': 'match stuff like: Sound.prototype.play = function() { … }'
-    'end': '(\\))'
-    'endCaptures':
-      '1':
-        'name': 'punctuation.definition.parameters.end.js'
-    'name': 'meta.function.prototype.js'
+    # [async] function [name](params)
+    # function* name(params) – generator function declaration
+    'begin': '(?=(?:\\basync\\b\\s*)?\\bfunction\\b)'
```
50Wliu: No need for `?:` non-capturing group because we're not capturing anything here anyway.

grammars/javascript.cson
```diff
@@ -281,240 +281,283 @@
...
+    'end': '(?<=})'
     'patterns': [
       {
-        'include': '#function-params'
+        'begin': '\\G'
+        'end': '(?<=\\))'
+        'name': 'meta.function.js'
```
50Wliu: Will this include the `async function()` part?

MaximSokolov: Yes
```
async function(...) {}
^-----------------^ - meta.function
```

grammars/javascript.cson
50Wliu: Are you sure?  I just tried this and it worked:
```js
var Sound = function() {};
myfunc = function() { console.log('hi') };
Sound.prototype.play = myfunc;
var sound = new Sound();
console.log(sound.play()); // --> hi
```

grammars/javascript.cson
```diff
@@ -281,240 +281,283 @@
...
-      '7':
-        'name': 'punctuation.definition.parameters.begin.js'
-    'comment': 'match stuff like: play = function() { … }'
-    'end': '(\\))'
-    'endCaptures':
-      '1':
-        'name': 'punctuation.definition.parameters.end.js'
-    'name': 'meta.function.js'
+    # "foo": function...
+    'begin': '(?=((\'[^\']*?\')|("[^"]*?"))\\s*:\\s*(\\basync\\b\\s*)?\\bfunction\\b)'
```
50Wliu: So from some testing with other grammars I think the following will also work: `((\'.*?\')|(".*?"))`

grammars/javascript.cson
MaximSokolov: It *can* be a function. Are you sure it always is used like that?

grammars/javascript.cson
```diff
@@ -1121,67 +1094,117 @@
...
       }
     ]
-  'function-params':
+  'function_params':
     'patterns': [
       {
-        'include': '#numbers'
+        'begin': '(\\()'
+        'beginCaptures':
+          '1':
```
50Wliu: 0 capture

grammars/javascript.cson
```diff
@@ -1121,67 +1094,117 @@
...
+              '1':
+                'name': 'keyword.operator.spread.js'
+              '2':
+                'name': 'variable.parameter.rest.function.js'
+          }
+          {
+            'include': '$self'
+          }
+          {
+            'match': '[a-zA-Z_$][a-zA-Z_$0-9]*'
```
50Wliu: Can this pattern go above the `$self` include?

grammars/javascript.cson
```diff
@@ -1121,67 +1094,117 @@
...
-        'match': '\\(|\\)'
-        'name': 'meta.brace.round.js'
+        'match': '(?:\\b(async)\\b\\s*)?\\b(function)\\b(?:\\s*(\\*))?'
+        'captures':
+          '1':
+            'name': 'storage.modifier.async.js'
+          '2':
+            'name': 'storage.type.function.js'
+          '3':
+            'name': 'storage.modifier.generator.asterisk.js'
```
50Wliu: I don't think the extra `.asterisk` is need here.

spec/javascript-spec.coffee
```diff
@@ -937,111 +927,161 @@ describe "Javascript grammar", ->
...
+      expect(tokens[26]).toEqual value: ',', scopes: ['source.js', 'meta.function.method.definition.js', 'meta.parameters.js', 'meta.delimiter.object.comma.js']
+      expect(tokens[28]).toEqual value: '"', scopes: ['source.js', 'meta.function.method.definition.js', 'meta.parameters.js', 'string.quoted.double.js', 'punctuation.definition.string.begin.js']
+      expect(tokens[33]).toEqual value: 'd', scopes: ['source.js', 'meta.function.method.definition.js', 'meta.parameters.js', 'variable.parameter.function.js']
+      expect(tokens[34]).toEqual value: ']', scopes: ['source.js', 'meta.function.method.definition.js', 'meta.parameters.js', 'meta.brace.square.js']
+      expect(tokens[35]).toEqual value: '}', scopes: ['source.js', 'meta.function.method.definition.js', 'meta.parameters.js', 'meta.brace.curly.js']
+      expect(tokens[36]).toEqual value: ')', scopes: ['source.js', 'meta.function.method.definition.js', 'meta.parameters.js', 'punctuation.definition.parameters.end.bracket.round.js']
+
+      {tokens} = grammar.tokenizeLine('write("){");')
```
50Wliu: Can we also test the semicolon?  The original point of this test was to make sure the string highlighting was correct.

spec/javascript-spec.coffee
```diff
@@ -937,111 +927,161 @@ describe "Javascript grammar", ->
...
+      expect(tokens[11]).toEqual value: '1', scopes: ['source.js', 'meta.function.js', 'meta.parameters.js', 'constant.numeric.decimal.js']
+
+      {tokens} = grammar.tokenizeLine('function callSomething(thing = this.something()) {}')
+      expect(tokens[4]).toEqual value: 'thing', scopes: ['source.js', 'meta.function.js', 'meta.parameters.js', 'variable.parameter.function.js']
+      expect(tokens[6]).toEqual value: '=', scopes: ['source.js', 'meta.function.js', 'meta.parameters.js', 'keyword.operator.assignment.js']
+      expect(tokens[8]).toEqual value: 'this', scopes: ['source.js', 'meta.function.js', 'meta.parameters.js', 'variable.language.js']
+      expect(tokens[9]).toEqual value: '.', scopes: ['source.js', 'meta.function.js', 'meta.parameters.js', 'meta.method-call.js', 'meta.delimiter.method.period.js']
+      expect(tokens[10]).toEqual value: 'something', scopes: ['source.js', 'meta.function.js', 'meta.parameters.js', 'meta.method-call.js', 'entity.name.function.js']
+
+    it "tokenizes rest parameter", ->
```
50Wliu: `it "tokenizes the rest parameter"`

spec/javascript-spec.coffee
```diff
@@ -1349,37 +1380,56 @@ describe "Javascript grammar", ->
...
-      expect(tokens[5]).toEqual value: ',', scopes: ['source.js', 'meta.function.js', 'meta.object.delimiter.js']
-      expect(tokens[7]).toEqual value: '//', scopes: ['source.js', 'meta.function.js', 'comment.line.double-slash.js', 'punctuation.definition.comment.js']
-      expect(tokens[8]).toEqual value: ' comment', scopes: ['source.js', 'meta.function.js', 'comment.line.double-slash.js']
+      expect(tokens[2]).toEqual value: '/*', scopes: ['source.js', 'meta.function.js', 'comment.block.js', 'punctuation.definition.comment.js']
+      expect(tokens[4]).toEqual value: '*/', scopes: ['source.js', 'meta.function.js', 'comment.block.js', 'punctuation.definition.comment.js']
+      expect(tokens[6]).toEqual value: 'foo', scopes: ['source.js', 'meta.function.js', 'entity.name.function.js']
+      expect(tokens[10]).toEqual value: '/*', scopes: ['source.js', 'comment.block.js', 'punctuation.definition.comment.js']
+      expect(tokens[12]).toEqual value: '*/', scopes: ['source.js', 'comment.block.js', 'punctuation.definition.comment.js']
+
+      {tokens} = grammar.tokenizeLine('x  => /* */ {}')
```
50Wliu: Extra space here after `x` (doesn't really matter though)

grammars/javascript.cson
50Wliu: Awesome.

grammars/javascript.cson
MaximSokolov: Specs fail on
```js
a.push("x" + y + ":function()");
a.push('x' + y + ':function()');
```

grammars/javascript.cson
MaximSokolov: e.g. `this` in `function callSomething(thing = this.something()) {}` should be matched first.
I am not sure what else can be inside function parameters, so put the pattern below `$self`

grammars/javascript.cson
50Wliu: Bleh.  Let's remove it then.  At the very least we'll be a bit more consistent.

grammars/javascript.cson
```diff
@@ -281,240 +281,283 @@
...
+        'include': '#function_body'
+      }
+      {
+        'include': '#comments'
       }
     ]
   }
   {
-    'begin': '\\b(constructor)\\s*(\\()'
+    'match': '(=>)'
```
50Wliu: In what circumstances do we hit this match?

grammars/javascript.cson
```diff
@@ -1111,67 +1084,117 @@
...
-        'name': 'variable.parameter.function.js'
+        'match': '([a-zA-Z_$][a-zA-Z_$0-9]*)(?=\\s*=>)'
+        'captures':
+          '0':
+            'name': 'meta.parameters.js'
+          '1':
+            'name': 'variable.parameter.function.js'
       }
-    ]
+  ]
```
50Wliu: This was outdented too much

grammars/javascript.cson
MaximSokolov: See 7165e972a7e692dd0

50Wliu: Dang.  We really need to try to make all these function matches newline-aware.  I'll create a new issue for that.

grammars/javascript.cson
```diff
@@ -281,240 +281,283 @@
         'name': 'variable.language.js'
   }
   {
-    'captures':
```
50Wliu: Why remove this capture?

grammars/javascript.cson
```diff
@@ -281,240 +281,283 @@
...
-    'end': '(\\))'
-    'endCaptures':
-      '1':
-        'name': 'punctuation.definition.parameters.end.js'
-    'name': 'meta.function.json.js'
+    # [param|(params)] => [expression|{statements}]
+    'begin': '''(?x)
+      (?=
+        (?<![A-Za-z0-9])
+        ((\\(([^\\(\\)]*)?\\))|[\\w$]+)
```
50Wliu: This doesn't catch invalid parameters like `0 => x * x` or `(0) => x * x` or `3d => x * y * z`.

grammars/javascript.cson
```diff
@@ -281,240 +281,283 @@
...
+        \\s*=>
+      )
+    '''
+    'end': '''(?x)
```
50Wliu: Can you guide me through this capture?  I'm not entirely sure I get what's going on here.

grammars/javascript.cson
```diff
@@ -281,240 +281,283 @@
...
+    # [.]foo = ... => ...
+    'begin': '''(?x)
+      (?=
+        (\\.)?[a-zA-Z_$][\\w$]*
+        \\s*(=)\\s*
+        ((\\(([^\\(\\)]*)?\\))|[\\w$]+)
+        \\s*=>
+      )
+    '''
+    'end': '''(?x)
```
50Wliu: Same with this one :smile:

grammars/javascript.cson
```diff
@@ -1111,65 +1084,115 @@
         ]
       }
     ]
-  'methods':
+  'arrow_functions':
```
50Wliu: So uh, this is never used.

50Wliu:
`function_params` doesn't catch this as invalid: `function hi(a b) {}`.  Not a priority for getting this PR merged though.

grammars/javascript.cson
MaximSokolov: `support.class` -> `variable.other.object` for `Sound` as it isn't a built-in class

grammars/javascript.cson
MaximSokolov:
```
x => /* */ x + x
|   |     ^ - (?=\\s*\\S)
|   ^ - (?!\\s*/\\*)
^ - (?!\\G[\\w$]+)

(...) => {...}
|       |     ^ - (?<=})
|       ^ - (?!\\s*{)
^ - (?!\\G\\()

x =>
 //comment
^ - (?!\\s*//)
{...}
```

grammars/javascript.cson
```diff
@@ -1111,65 +1084,102 @@
...
         'endCaptures':
-          '1':
-            'name': 'punctuation.definition.parameters.end.js'
+          '0':
+            'name': 'punctuation.definition.parameters.end.bracket.round.js'
+        'name': 'meta.parameters.js'
         'patterns': [
           {
-            'include': '#function-params'
+            'match': '(\\.\\.\\.)([a-zA-Z_$][a-zA-Z_$0-9]*)'
```
50Wliu: Second match can be shortened to `[\\w$]*`

grammars/javascript.cson
```diff
@@ -1111,65 +1084,102 @@
...
+                'name': 'keyword.operator.spread.js'
+              '2':
+                'name': 'variable.parameter.rest.function.js'
+          }
+          {
+            'include': '$self'
           }
           {
-            'include': '#strings'
+            'match': '[a-zA-Z_$][a-zA-Z_$0-9]*'
```
50Wliu: Same here

grammars/javascript.cson
50Wliu: Thanks!

50Wliu:
`function_params` is also missing `(01) => 1 + 2` or `function test(1) {}`.

50Wliu:
Regression time:
```js
function test()
{  //punctuation.definition.function.body.end.bracket.curly.js
    if(something)
    { // meta.brace.curly.js
        
    } // punctuation.definition.function.body.end.bracket.curly.js 
} // meta.brace.curly.js
```

Same with arrow functions.  We need to wrap curly bracket matching in a begin/end instead of a `|` match.

Also, in the case of `Git.Repository.open()`, should `Repository` really be tokenized as `variable.other.property.js`?  Seems to me like it's an object just like `Git`.

50Wliu:
Here's another problem I found: typing the following code (same as the above example) incorrectly marks the `if` as a function instead of a control keyword.  Adding/removing a space to try to get the grammar to realize its mistake doesn't work, but copy/pasting the code elsewhere or reloading the file does.

```js
function test()
{
    if(something)  // entity.name.function.js!!!
    {
        
    }
}
```

MaximSokolov:
> `function_params` is also missing (01) => 1 + 2 or function test(1) {}

Numbers cannot be matched as illegal because of default parameters:
```js
function multiply(a, b = 1){}
```

grammars/javascript.cson
```diff
@@ -281,240 +281,283 @@
...
-    'name': 'meta.function.prototype.js'
+    # [async] function [name](params)
+    # function* name(params) – generator function declaration
+    'begin': '(?=(\\basync\\b\\s*)?\\bfunction\\b)'
+    'end': '(?<=})'
     'patterns': [
       {
-        'include': '#function-params'
+        'begin': '\\G'
+        'end': '(?<=\\))'
```
MaximSokolov: @50Wliu any ideas why it's happening? 

thimpq3r7n

```js
function()

async function(){}
```

50Wliu: Looking into it...what's the highlighted scope that's flickering?

50Wliu: Never mind, got it: `meta.function.js`.  Not sure why though.  It only seems to happen when there's an odd number of spaces.

MaximSokolov: The scope is `meta.function` It cause problem with wrong `if` highlighting:
```js
function test()
{
    if(something)  // entity.name.function.js!!!
    {

    }
}
```

50Wliu: Moving the function_body include above this match appears to fix it.  For safety I would arrange the patterns like this:
* comments
* function_body
* This one

MaximSokolov:
Added support for:
```js
{
  foo: p => {}
  "foo": p => {}
  'foo': p => {}
}
```

50Wliu:
Looks good to me - can you just address the two style comments above?
https://github.com/atom/language-javascript/pull/312#discussion-diff-52385139
and
https://github.com/atom/language-javascript/pull/312#discussion-diff-52385190

I wasn't able to find any glaring bugs through some testing, so I think this is ready to :ship: and get some more extensive testing on Atom master :).

MaximSokolov:
done

PR #313, opened 2016-01-25
Add jsx to file types

calebmer:
No JSX package on atom compares this one to parsing JavaScript. I'd like to just not use any of these packages in my `.jsx` files and just have `language-javascript` render them.

### DIFF
--- a/grammars/javascript.cson
+++ b/grammars/javascript.cson
@@ -13,6 +13,7 @@
   'pjs'
   'xsjs'
   'xsjslib'
+  'jsx'
 ]
 'firstLineMatch': '^#!.*\\b(node|iojs|JavaScript)'
 'name': 'JavaScript'


lee-dohm:
You can add this yourself by adding a `customFileTypes` declaration to your `config.cson`:

```
'*':
  'core':
    'customFileTypes':
      'source.js': [
        'jsx'
      ]
```

This way you get what you want and it won't break other people using JSX packages :grinning:

calebmer:
That sounds sensible, but including JSX seems like a reasonable default
include to me.

lee-dohm:
I didn't close this for now, going to bounce the idea off the team.

50Wliu:
linguist, the parser GitHub uses to highlight files on its website, used to include `.jsx` as JavaScript, which led to pretty bad experiences (see #220 or github/linguist#2624 as an example).  Since language-javascript doesn't explicitly support JSX, that leads to subpar tokenization at best.

I think going with the `customFileTypes` option is your best bet here other than waiting for the JSX packages to improve.  Out of curiousity, have you tried `language-babel`?  That's what linguist uses now.

PR #319, opened 2016-02-01
Don't scope decimals as illegal; tokenize decimal marks, illegal identifiers

### LINKED ISSUE #318: Inconsistent scoping of decimals
I noticed some bugs on decimal numbers:

c

```javascript
0.9 == .9

9.0 == 9.
```

a

b

On the first line, `.9` shouldn't be illegal: [Floating-point literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Grammar_and_types#Floating-point_literals).

Also, periods in numbers should have their own scope IMO, different from `constant.numeric`, the same way plus and minus signs are `keyword.operator`.

MaximSokolov:
```js
// octal
015, 0010, 0o77, 0O10

// hexadecimal
0xFF, 0xff, 0x1123, 0X00111, 0XF1A7

// binary
0b11, 0b0011, 0b11,                 

// decimal
// [(+|-)][digits][.digits][(E|e)[(+|-)]digits]
1E+12
3.1415926
.123456789
3.1E+12
.1e-23

1
1.
.1
1.1
1E+3
1.E3
.1E+3
1.1E+3

// illegal identifiers
1abc
123abc
1$
123$abc
.1a
obj.1
obj.1a
1$.123
```

Fixes #318

### DIFF
--- a/grammars/javascript.cson
+++ b/grammars/javascript.cson
@@ -1211,15 +1211,13 @@
     'patterns': [
       {
         'comment': '`prop` in `obj.prop`, `func().prop`'
-        'match': '(\\.)\\s*(?:(\\b\\d\\w*)|([A-Z][A-Z0-9_$]*\\b\\$*)|([\\w$]*))'
+        'match': '(\\.)\\s*(?:([A-Z][A-Z0-9_$]*\\b\\$*)|(\\$*[a-zA-Z_$][\\w$]*))'
         'captures':
           '1':
             'name': 'meta.delimiter.property.period.js'
           '2':
-            'name': 'invalid.illegal.js'
-          '3':
             'name': 'constant.other.property.js'
-          '4':
+          '3':
             'name': 'variable.other.property.js'
       }
     ]
--- a/grammars/javascript.cson
+++ b/grammars/javascript.cson
@@ -918,8 +918,35 @@
   'numbers':
     'patterns': [
       {
-        'match': '\\b(?<!\\$|_)((0(x|X)[0-9a-fA-F]+)|(0(b|B)[01]+)|(0(o|O)[0-7]+)|([0-9]+(\\.[0-9]+)?)((e|E)[+-]?[0-9]+)?)\\b'
-        'name': 'constant.numeric.js'
+        'match': '\\b(?<!\\$|_)0(x|X)[0-9a-fA-F]+'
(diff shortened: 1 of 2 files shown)


spec/javascript-spec.coffee
```diff
@@ -1220,11 +1224,6 @@ describe "Javascript grammar", ->
       expect(tokens[3]).toEqual value: '.', scopes: ['source.js', 'meta.delimiter.property.period.js']
       expect(tokens[4]).toEqual value: 'b', scopes: ['source.js', 'variable.other.property.js']
 
-      {tokens} = grammar.tokenizeLine('a.123illegal')
```
50Wliu: Is this impossible to test for now?  Can't we just give numbers higher priority?

grammars/javascript.cson
```diff
@@ -918,8 +918,35 @@
...
+              (?:\\b[0-9]+)
+            )
+            ([eE][+-]?[0-9]+)?
+          )
+        '''
+        'captures':
+          '0':
+            'name': 'constant.numeric.decimal.js'
+          '1':
+            'name': 'meta.delimiter.decimal-mark.period.js'
```
50Wliu: I'm not sold on `decimal-mark` - it sounds really weird to me.  Maybe just `decimal`.

MaximSokolov:
updated

50Wliu:
Awesome!  @chbk would you be so kind to test this as well? :stuck_out_tongue:

grammars/javascript.cson
```diff
@@ -918,8 +922,42 @@
...
+        'name': 'constant.numeric.binary.js'
+      }
+      {
+        'match': '\\b(?<!\\$)0(o|O)?[0-7]+\\b(?!\\$)'
+        'name': 'constant.numeric.octal.js'
+      }
+      {
+        'match': '''(?x)
+          (?<!\\$)(?:
+            (?:\\b[0-9]+(\\.)[0-9]+[eE][+-]?[0-9]+\\b)| # 1.1E+3
```
50Wliu: `1.E3` is also possible...it doesn't look like you're catching that right now

chbk:
Tested and working :+1:

50Wliu:
Thanks for testing @chbk!

PR #321, opened 2016-02-02
Fix function/method calls being superseded

### LINKED ISSUE #294: Function-call highlights being superceded

### LINKED ISSUE #320: [ES6] Syntax highlight error
This is a pretty minor one, but still:

---

Notice the color of the selected bracket.

MaximSokolov:
Continuation of #295

Fixes #294
Fixes #320
Replaces and closes #295


grammars/javascript.cson
```diff
@@ -617,16 +617,34 @@
...
-  }
-  {
-    'match': '\\b(\\.)(assert|clear|debug|error|info|log|profile|profileEnd|time|timeEnd|warn)(?=\\()'
-    'captures':
-      '1':
-        'name': 'meta.delimiter.method.period.js'
-      '2':
-        'name': 'support.function.js.console'
+    # console.log(arg1, "arg2", [...])
+    'begin': '\\bconsole'
```
50Wliu: What if we have `consolea.log()`?

grammars/javascript.cson
```diff
@@ -617,16 +617,34 @@
...
+    'beginCaptures':
+      '0':
+        'name': 'entity.name.type.object.console.js'
+    'end': '(?!\\G)'
```
50Wliu: Why do we need a `(?!\\G)` here?  Does a `(?<=\\)` not suffice?

grammars/javascript.cson
```diff
@@ -617,16 +617,34 @@
...
-    'captures':
-      '1':
-        'name': 'meta.delimiter.method.period.js'
-      '2':
-        'name': 'support.function.js.console'
+    # console.log(arg1, "arg2", [...])
+    'begin': '\\bconsole'
+    'beginCaptures':
+      '0':
+        'name': 'entity.name.type.object.console.js'
```
50Wliu: :+1: Finally fixed that switched naming

grammars/javascript.cson
```diff
@@ -1188,16 +1182,73 @@
...
+                  tabIndex|insertRow|insertBefore|insertCell|insertData|item|open|deleteRow|deleteCell|
+                  deleteCaption|deleteTHead|deleteTFoot|deleteData|focus|write|writeln|add|appendChild|
+                  appendData|reset|replaceChild|replaceData|move|moveNamedItem|moveChild|moveAttribute|
+                  moveAttributeNode|getNamedItem|getElementsByName|getElementsByTagName|getElementById|
+                  getAttribute|getAttributeNode|blur)\\b
+                '''
+                'name': 'support.function.dom.js'
+              }
+              {
+                'match': ".+"
```
50Wliu: This seems dangerous...we should make sure the function name is valid.

grammars/javascript.cson
```diff
@@ -1212,16 +1263,29 @@
...
+              {
+                'match': '''(?x)
+                  \\b(isNaN|isFinite|eval|uneval|parseInt|parseFloat|decodeURI|
+                  decodeURIComponent|encodeURI|encodeURIComponent|escape|unescape|
+                  require|set(Interval|Timeout)|clear(Interval|Timeout))\\b
+                '''
+                'name': 'support.function.js'
+              }
+              {
+                'match': ".+"
```
50Wliu: Ditto.

grammars/javascript.cson
MaximSokolov: It also matches the `console ` keyword without methods

grammars/javascript.cson
```diff
@@ -617,16 +617,30 @@
...
+    # console
+    # console.log(arg1, "arg2", [...])
+    'begin': '\\bconsole\\b'
+    'beginCaptures':
+      '0':
+        'name': 'entity.name.type.object.console.js'
+    'end': '(?!\\G)'
+    'patterns': [
+      {
+        'begin': '\\b(\\.)(assert|clear|debug|error|info|log|profile|profileEnd|time|timeEnd|warn)\\s*(?=\\()'
```
50Wliu: `\\s*(\\.)\\s*`
Have to test for spaces here as well.  I don't think we need the `\\b` anymore if we test for spaces.

grammars/javascript.cson
```diff
@@ -1185,25 +1175,94 @@
...
       }
     ]
-  'method_call':
+  'method_calls':
     'patterns': [
       {
-        'comment': '.methodCall(arg1, "arg2", [...])'
-        'begin': '(\\.)([a-zA-Z_$][\\w$]*)\\s*(\\()'
+        # .methodCall(arg1, "arg2", [...])
+        'begin': '(\\.)([\\w$]+)\\s*(?=\\()'
```
50Wliu: Nice catch on removing the duplicate `a-zA-Z_$` block.

50Wliu: Actually...we need that back since `\w` includes numbers, which can't start a method call.

grammars/javascript.cson
```diff
@@ -1185,25 +1175,94 @@
...
+                  getAttribute|getAttributeNode|blur)\\b
+                '''
+                'name': 'support.function.dom.js'
+              }
+              {
+                'match': "[a-zA-Z_$][\\w$]*"
+                'name': 'entity.name.function.js'
+              }
+              {
+                'match': '[\\d][\\w$]*'
```
50Wliu: This can be `\\d+[\\w$]*`

grammars/javascript.cson
```diff
@@ -1212,22 +1271,32 @@
...
+                  require|set(Interval|Timeout)|clear(Interval|Timeout))\\b
+                '''
+                'name': 'support.function.js'
+              }
+              {
+                'match': "[a-zA-Z_$][\\w$]*"
+                'name': 'entity.name.function.js'
+              }
+              {
+                'match': '[\\d][\\w$]*'
```
50Wliu: Same here.

50Wliu:
Overall looks pretty good.  Left a few small comments.

50Wliu:
Also, it would be great if we could tokenize this properly:
```js
console
.
log(); // <-- meta.function-call, no support.function.console.js
```
or
```js
console.
log();
```
etc.  Not just limited to `console`, for example:
```js
something
.
hi();
```

grammars/javascript.cson
MaximSokolov: It allows to match arguments (currently not tagged) even if function name is illegal:

50Wliu:
@MaximSokolov Is this ready for merge?  It looks good to me.

MaximSokolov:
:ship:

PR #311, opened 2016-01-24
Tokenize built-in variables with properties

MaximSokolov:
See [#309 (comment)](https://github.com/atom/language-javascript/issues/309#issuecomment-174328919)

PR #327, opened 2016-02-24
Improve objects, properties

### LINKED ISSUE #326: Mark `CONSTANT` properties as constants
For example, when I have a constant which is an object, it becomes a regular `variable.other.object.js` and loses the constant scope `constant.other.js` and corresponding coloring:

image

MaximSokolov:
Closes #326 (actually was fixed by #311)

### DIFF
--- a/spec/javascript-spec.coffee
+++ b/spec/javascript-spec.coffee
@@ -1278,7 +1278,7 @@ describe "Javascript grammar", ->
 
       {tokens} = grammar.tokenizeLine('a.123illegal')
       expect(tokens[0]).toEqual value: 'a', scopes: ['source.js']
-      expect(tokens[1]).toEqual value: '.', scopes: ['source.js', 'meta.delimiter.method.period.js']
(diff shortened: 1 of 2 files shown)


spec/javascript-spec.coffee
```diff
@@ -1270,7 +1278,7 @@ describe "Javascript grammar", ->
 
       {tokens} = grammar.tokenizeLine('a.123illegal')
       expect(tokens[0]).toEqual value: 'a', scopes: ['source.js']
-      expect(tokens[1]).toEqual value: '.', scopes: ['source.js', 'meta.delimiter.property.period.js']
+      expect(tokens[1]).toEqual value: '.', scopes: ['source.js', 'meta.delimiter.method.period.js']
```
50Wliu: This should still be `.property`

MaximSokolov: fixed

kartikanand:
This fixes #363  as well! :smile:

PR #331, opened 2016-02-29
Fix regex in arrow function

### LINKED ISSUE #315: Highlighting broken when regex contains speech marks
I first noticed this problem on github.com itself, but have noticed it also affects the atom highlighting (I believe the same scheme is used? At least that is what github support told me).

Here is the offending code, which you can see is broken here on the site itself!

```js
const isCssImage = val => /^url\(["']?http/.test(val);

function something() {
  console.log("WTF, highlighting completely screwed");
}
```

And here is a screenshot of it broken in Atom:

screen shot 2016-01-28 at 22 55 19

### LINKED ISSUE #329: Regex and Arrow Function
When I have a bare regex immediately after the `=>` of an arrow function, the regex doesn't seem to be recognized.

regex

MaximSokolov:
Fixes #329
Fixes #315

### DIFF
--- a/grammars/javascript.cson
+++ b/grammars/javascript.cson
@@ -901,7 +901,7 @@
     'name': 'constant.language.js'
   }
   {
-    'begin': '(?<=[\\[=(?:+,!]|^|return|&&|\\|\\|)\\s*(/)(?![/*+{}?])(?=.*/)'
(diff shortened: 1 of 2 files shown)


thomasjo:
Looks good to me!

PR #335, opened 2016-03-10
Add jsx file extension to fileTypes

cusspvz:
`language-js` parses correctly `jsx` files (es6) but it isn't defined on fileTypes.

Turns out that I have to always change the language type to `Javascript` for `jsx` file types, which is very painful specially for those who use `.jsx` for **react components** files.

### DIFF
--- a/grammars/javascript.cson
+++ b/grammars/javascript.cson
@@ -2,6 +2,7 @@
 'fileTypes': [
   'js'
   '_js'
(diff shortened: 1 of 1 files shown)


cusspvz:
For those who came from google, I've installed [language-babel](https://atom.io/packages/language-babel).

@50Wliu thanks for your fast response, appreciated.

Issue #345, opened 2016-03-29
Make CommonJS require const


javierjmx:
insert:
`const koa = require('koa');`

instead of:
`var koa = require('koa');`

Reference: http://airbnb.io/javascript/#13.1

50Wliu:
Fixed by #346.

PR #346, opened 2016-03-29
Declare Required Modules As Const Instead Of Var

javierjmx:
Part of #345 

Reference: http://airbnb.io/javascript/#13.1


50Wliu:
Thanks!

PR #349, opened 2016-04-07
Add SIMD highlighting.

CentricStorm:
Highlight SIMD object and data types from the [SIMD API](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SIMD).

### DIFF
--- a/grammars/javascript.cson
+++ b/grammars/javascript.cson
@@ -724,6 +724,41 @@
     ]
   }
   {
+    # SIMD object & data types
+    'begin': '\\bSIMD\\b'
+    'beginCaptures':
+      '0':
+        'name': 'entity.name.type.object.simd.js'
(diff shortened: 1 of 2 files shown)


thomasjo:
I'm a bit torn on this, since the SIMD proposal is still in [draft stage](https://tc39.github.io/ecmascript_simd/) in TC39. We have to maintain the language grammar, and since this can change at any time, or even be rejected, I propose that we hold off on merging this until at least v1.0.0 of the SIMD specification.

/cc @atom/feedback

50Wliu:
Agreed on waiting until the proposal becomes stable.

CentricStorm:
Closing as "[SIMD.js has been taken out of active development](https://github.com/tc39/ecmascript_simd)".

Issue #344, opened 2016-03-29
Wrong scope on $this and $default


chbk:
Title says it all: `this` in `$this` and `abc$this` is `variable.language.js`. I haven't tested all keywords but it happens with `default` and `new` too.

50Wliu:
Thanks for the report.

PR #242, opened 2015-10-23
Support for properties, function/method calls

MaximSokolov:
Adds support for properties, function calls, method calls.

Fixes #37, fixes #241
Replaces: closes #217, closes #146
Refs #137

PR #351, opened 2016-04-07
Update Math highlighting.

CentricStorm:
- Update [Math](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Math) methods and properties to ECMAScript 2015.
- Separate Math methods/properties from others so that they are only highlighted when preceded by the Math object (e.g. highlight `Math.random()` but not `foo.random()`).
- Add Math unit tests.

### DIFF
--- a/grammars/javascript.cson
+++ b/grammars/javascript.cson
@@ -724,6 +724,41 @@
     ]
   }
   {
+    # Math
+    'begin': '\\bMath\\b'
+    'beginCaptures':
+      '0':
+        'name': 'support.class.js'
+    'end': '(?!\\G)'
+    'patterns': [
+      {
+        # Math.random()
+        'begin': '\\s*(\\.)\\s*(abs|acos|acosh|asin|asinh|atan|atan2|atanh|cbrt|ceil|clz32|cos|cosh|exp|expm1|floor|fround|hypot|imul|log|log10|log1p|log2|max|min|pow|random|round|sign|sin|sinh|sqrt|tan|tanh|trunc)\\s*(?=\\()'
+        'beginCaptures':
+          '1':
+            'name': 'meta.delimiter.method.period.js'
+          '2':
+            'name': 'support.function.math.js'
+        'end': '(?<=\\))'
+        'name': 'meta.method-call.js'
+        'patterns': [
+          {
+            'include': '#arguments'
+          }
+        ]
+      }
+      {
+        # Math.PI
+        'match': '\\s*(\\.)\\s*(E|LN10|LN2|LOG10E|LOG2E|PI|SQRT1_2|SQRT2)\\b'
+        'captures':
+          '1':
+            'name': 'meta.delimiter.property.period.js'
+          '2':
+            'name': 'support.constant.property.math.js'
+      }
+    ]
+  }
+  {
     'include': '#strings'
   }
   {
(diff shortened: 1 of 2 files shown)


grammars/javascript.cson
```diff
@@ -724,6 +724,41 @@
     ]
   }
   {
+    # Math
+    'begin': '\\bMath\\b'
```
50Wliu: This will catch `Math$hi` I believe.

grammars/javascript.cson
```diff
@@ -724,6 +724,41 @@
...
+        'name': 'meta.method-call.js'
+        'patterns': [
+          {
+            'include': '#arguments'
+          }
+        ]
+      }
+      {
+        # Math.PI
+        'match': '\\s*(\\.)\\s*(E|LN10|LN2|LOG10E|LOG2E|PI|SQRT1_2|SQRT2)\\b'
```
50Wliu: These should (I feel) be marked as a constant.
You should also make sure there's no parentheses following it, or any non-word characters like `$`, `.`, etc.

CentricStorm: What scope should be used instead of `support.property.math.js`? The old scope was `support.constant.js`.

PI is both a constant _and_ a property of the Math object, so `support.property.constant.math.js`?

grammars/javascript.cson
CentricStorm: `\b` matches word boundaries; it doesn't match $ (tested in Atom 1.7.0). Both JavaScript and Oniguruma regular expressions classify `\w` characters as `[A-Za-z0-9_]`.

`\\bMath\\b` is consistent with `\\bconsole\\b` which is already in the code.

grammars/javascript.cson
50Wliu: /cc @MaximSokolov

MaximSokolov: :+1: for `support.constant.property.math.js`
According to [naming conventions](http://manual.macromates.com/en/language_grammars#naming_conventions) this is `support.constant`. I added `property` subscope in #242 to distinguish variables, I think `property` also should be added to reserved constants for consistent.

CentricStorm:
Changed Math constant scope to `support.constant.property.math.js`.

Also note that the old scope for Math was `support.class.js` which has been changed to `entity.name.type.object.math.js`. This is more accurate because the Math object is not a class (or a constructor), and is consistent with `entity.name.type.object.console.js`

MaximSokolov:
> This is more accurate because the Math object is not a class (or a constructor), and is consistent with `entity.name.type.object.console.js`

Maybe "class" is not accurate term for `Math`, but there is no standard scope for reserved objects (`support.type` is for things like `size_t`, `support.other.object` won't be highlighted in syntax themes which color only `support.class`), so I suggest keep `support.class` for `Math`

`console` should be `support.class` as well because it's a reserved object, not a “placeholder”. I didn't change it, as renaming `entity.name` -> `support.class` will cause highlighting changes. People don't really like breaking changes.

CentricStorm:
> Maybe "class" is not accurate term for `Math`, but there is no standard scope for reserved objects (`support.type` is for things like `size_t`, `support.other.object` won't be highlighted in syntax themes which color only `support.class`), so I suggest keep `support.class` for `Math`

Changed Math scope back to `support.class.js`.

grammars/javascript.cson
50Wliu: Sorry, what I meant was that it will successfully match `Math` as `support.class.js` and attempt to begin matching the subgroups.

CentricStorm: This is what the commit currently it looks like in Atom 1.7.0:
math-highlight-example
Same behaviour as `console` and others.

50Wliu: Yep, and these bugs are sorta-kinda documented at #344 :wink:.

I guess we'll just have to fix these issues in another PR.

50Wliu:
Thanks!

PR #352, opened 2016-04-10
Added support for Unity3d WebGL native plugin types

elaberge:
As per http://docs.unity3d.com/Manual/webgl-interactingwithbrowserscripting.html
and
https://unity3d.com/unity/whats-new/unity-5.2


50Wliu:
Thanks!

PR #355, opened 2016-04-13
Add support for matching console as is, fix for issue 354

kartikanand:
The current grammar expects `console` to be always used as `console.(log|error|etc.)()` function call. Although, it can also be passed to other functions, assigned etc. This is causing syntax issue as referenced in issue 354. This fix involves, matching `console` which is not directly followed by a period `.`

### DIFF
--- a/grammars/javascript.cson
+++ b/grammars/javascript.cson
@@ -709,11 +709,6 @@
     'end': '(?!\\G)'
     'patterns': [
       {
-        # Match only console. When used as an object directly
-        # For eg. foo(console)
-        'match': '\\s*(?!\\.)'
-      }
-      {
         'begin': '\\s*(\\.)\\s*(assert|clear|debug|error|info|log|profile|profileEnd|time|timeEnd|warn)\\s*(?=\\()'
         'beginCaptures':
           '1':
@@ -727,6 +722,11 @@
             'include': '#arguments'
           }
         ]
+      }
+      {
+        # Match only console. When used as an object directly
+        # For eg. foo(console)
+        'match': '\\s*(?!\\.)'
       }
     ]
   }
--- a/grammars/javascript.cson
+++ b/grammars/javascript.cson
@@ -709,6 +709,11 @@
     'end': '(?!\\G)'
     'patterns': [
       {
+        # Match only console. When used as an object directly
+        # For eg. foo(console)
+        'match': '\\s*(?!\\.)'
+      }
+      {
         'begin': '\\s*(\\.)\\s*(assert|clear|debug|error|info|log|profile|profileEnd|time|timeEnd|warn)\\s*(?=\\()'
         'beginCaptures':
           '1':


50Wliu:
/cc @MaximSokolov - any idea why the `(?!\\G)` isn't enough by itself to stop the matching?
`console;` also doesn't work properly - the semicolon is still treated as part of the console match (but anything after that isn't).

kartikanand:
@50Wliu  isn't `(?!\\G)` telling it to not stop just after matching `console` ? Since, when `console` is matched, `\G` would equal the position just after `console`, and by saying `end: (?!\\G)`, we're effectively saying, that end should be when the current position is not after the current match.

kartikanand:
Ok, so after discussions with Wictor on StackOverflow http://stackoverflow.com/questions/36590191/what-does-the-regex-g-do-in-language-grammars/36597991?noredirect=1#comment60840799_36597991  , it seems the issue is due to adding keywords after `log`, which the current behavior only expects to be paranthesis `()`. So I'm deleting the current pull request, and will be making a new one. In the comment, Wictor has only added `bind`, but I guess it can also be `call` and `apply`.

CentricStorm:
> cc @MaximSokolov - any idea why the `(?!\\G)` isn't enough by itself to stop the matching?
> `console;` also doesn't work properly - the semicolon is still treated as part of the console match (but anything after that isn't).

`(?!\\G)` matches the first character that is not followed by the last character of the previous match, so normally it would match the 'e' in `console`, but...

...the syntax parser *can only match a given character once*, and `console` has *already been matched*, so instead the parser simply matches any character after `console`.

50Wliu:
Makes sense now - thanks @kartikanand and @CentricStorm (I'm still learning about `\G`).

kartikanand:
@CentricStorm  One doubt! As you said "(?!\\G) matches the first character that is not followed by the last character of the previous match", so it should match any character following `console`.

So, for the following `foo(console)`, the end should match the closing paranthesis `)`, and the editor should stop there. Then why the inconsistent highlighting in the following? 

screenshot from 2016-04-14 18 25 04

CentricStorm:
> So, for the following `foo(console)`, the end should match the closing paranthesis `)`, and the editor should stop there. Then why the inconsistent highlighting in the following?

In that example the `(` is matched by the syntax parser normally, but the `)` is incorrectly matched by the `console` end pattern, so the parser gets stuck inside the `meta.arguments.js` scope all the way to the end of the file.

kartikanand:
@CentricStorm  Sorry I deleted my comments because I thought it was a stupid thing to ask! Thanks for answering and making it clearer!

kartikanand:
@MaximSokolov  @CentricStorm  Aren't look-around supposed to only assert a character and not capture it.

Let me explain:

I created the following dummy grammar:

```
'scopeName': 'source.js'
'fileTypes': [
  'js'
]
'firstLineMatch': '^#!.*\\b(node|iojs|JavaScript)'
'name': 'JavaScript'
'patterns': [
  {
    'begin': '(?<=#)\\w+'
    'end': '(?=#)'
    'name': 'comment'
  }
  {
    'match': '#'
    'name': 'token'
  }
  {
    'begin': '\\bconsole\\b'
    'name': 'console'
    'end': '(?!\\G)'
  }
  {
    'match': '\\('
    'name': 'bracket open'
  }
  {
    'match': '\\)'
    'name': 'bracket close'
  }
]
```

Then ran the following code on it:

```
var fm = require('first-mate');
var gr = fm.GrammarRegistry;
var r = new gr();
var g = r.loadGrammarSync('../language-javascript/grammars/javascript.cson');

var  o = g.tokenizeLine('#comment#');
var tok = r.decodeTokens(o.line, o.tags);
console.log(tok);

o = g.tokenizeLine('(console)');
tok = r.decodeTokens(o.line, o.tags);
console.log(tok);
```

For which I got the following output:

```
[ { value: '#', scopes: [ 'source.js', 'token' ] },
  { value: 'comment', scopes: [ 'source.js', 'comment' ] },
  { value: '#', scopes: [ 'source.js', 'token' ] } ]
[ { value: '(', scopes: [ 'source.js', 'bracket open' ] },
  { value: 'console', scopes: [ 'source.js', 'console' ] },
  { value: ')', scopes: [ 'source.js', 'console' ] } ]
```

As you can see, the `#` tokens are not captured by the comment grammar, but `)` is captured by `console` grammar, why is this behavior?

kartikanand:
Also see this link -> https://regex101.com/r/fL7gQ4/1

PR #358, opened 2016-04-14
Fix console, Math matching

### LINKED ISSUE #354: Console keyword causing problems
I came here from atom/language-html.

Basically the `console` keyword used within `<script>` tags is causing problems with syntax highlighting.

error

The problem is due to using the keyword console directly.

See below, when I replace console with konsole, it works:

working

Link to lanugage-html issue: https://github.com/atom/language-html/issues/111

MaximSokolov:
Fixes #354
Maybe @kartikanand knowns a better solution

Refs #355
Refs #344 (wrong scopes with `$`)

### DIFF
--- a/grammars/javascript.cson
+++ b/grammars/javascript.cson
@@ -702,7 +702,7 @@
   {
     # console
     # console.log(arg1, "arg2", [...])
-    'begin': '\\bconsole\\b(?!\\$)'
+    'begin': '(?<![\\w$])console(?![\\w$]|\\s*:)'
     'beginCaptures':
       '0':
         'name': 'entity.name.type.object.console.js'
@@ -727,7 +727,7 @@
   }
   {
     # Math
-    'begin': '\\bMath\\b'
+    'begin': '(?<![\\w$])Math(?![\\w$]|\\s*:)'
     'beginCaptures':
       '0':
         'name': 'support.class.js'
@@ -902,7 +902,7 @@
         'name': 'keyword.other.js'
   }
   {
-    'match': '(?<!\\$)\\b(Anchor|Applet|Area|Array|Boolean|Button|Checkbox|Date|document|event|FileUpload|Form|Frame|Function|Hidden|History|Image|JavaArray|JavaClass|JavaObject|JavaPackage|java|Layer|Link|Location|Map|Math|MimeType|Number|navigator|netscape|Object|Option|Packages|Password|Plugin|Radio|RegExp|Reset|Select|Set|String|Style|Submit|Symbol|screen|sun|Text|Textarea|WeakMap|WeakSet|window|XMLHttpRequest)\\b'
+    'match': '(?<!\\$)\\b(Anchor|Applet|Area|Array|Boolean|Button|Checkbox|Date|document|event|FileUpload|Form|Frame|Function|Hidden|History|Image|JavaArray|JavaClass|JavaObject|JavaPackage|java|Layer|Link|Location|Map|MimeType|Number|navigator|netscape|Object|Option|Packages|Password|Plugin|Radio|RegExp|Reset|Select|Set|String|Style|Submit|Symbol|screen|sun|Text|Textarea|WeakMap|WeakSet|window|XMLHttpRequest)\\b'
     'name': 'support.class.js'
   }
   {
--- a/grammars/javascript.cson
+++ b/grammars/javascript.cson
@@ -711,7 +711,7 @@
         (?! \\s*(\\.)\\s*
           (assert|clear|debug|error|info|log|profile|profileEnd|time|timeEnd|warn)
           \\s*\\(
-        ) \\S
+        ) \\s*\\S
       )
     '''
     'patterns': [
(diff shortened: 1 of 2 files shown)


kartikanand:
Yours is a way better and clean solution. One thing I don't understand, why the need for `(?!\\$)` in `begin` ?

MaximSokolov:
Because of incorrect word boundaries(issue #344 ):
```js
console$
```
However `(?!\\$)'` doesn't work for `Math` for some reason
```coffee
'begin': '\\bMath\\b(?!\\$)'
```
```js
Math$
```

CentricStorm:
Still more problems to fix:

console-math-problems

grammars/javascript.cson
```diff
@@ -702,11 +702,11 @@
   {
     # console
     # console.log(arg1, "arg2", [...])
-    'begin': '\\bconsole\\b'
+    'begin': '\\bconsole\\b(?!\\$)'
```
kartikanand: Maybe we can do something like -

`'begin': '(?<!\\$)\\bconsole\\b(?!\\$|//|/\*)'`

grammars/javascript.cson
```diff
@@ -702,11 +702,11 @@
...
     'beginCaptures':
       '0':
         'name': 'entity.name.type.object.console.js'
-    'end': '(?!\\G)'
+    'end': '(?=(?!\\s*\\.)\\S)'
     'patterns': [
       {
         'begin': '\\s*(\\.)\\s*(assert|clear|debug|error|info|log|profile|profileEnd|time|timeEnd|warn)\\s*(?=\\()'
```
kartikanand: Here we can capture all properties?

`'begin': '\\s*(\\.)\\s*(\w+)\\s*(?=\\()'`

kartikanand: Here we can capture all properties?

'begin': '\\s*(\\.)\\s*(\w+)\\s*(?=\\()'

kartikanand: Here we can capture all properties?

`'begin': '\s(\.)\s(\w+)\s*(?=\()'`

MaximSokolov:
> But for some reason it's not working for new lines, maybe begin doesn't take newlines into account: 
```js
console
: 'asd'
```

The tokenizer works line by line, so nothing you can do about this issue

> Should not be highlighted: 
```js
console//
console/**/
```

Looks legit to me
image

grammars/javascript.cson
```diff
@@ -702,14 +702,21 @@
...
+        (?! \\s*(\\.)\\s*
+          (assert|clear|debug|error|info|log|profile|profileEnd|time|timeEnd|warn)
+          \\s*\\(
+        ) \\S
+      )
+    '''
     'patterns': [
       {
-        'begin': '\\s*(\\.)\\s*(assert|clear|debug|error|info|log|profile|profileEnd|time|timeEnd|warn)\\s*(?=\\()'
+        'begin': '\\s*(\\.)\\s*(\\w+)\\s*(?=\\()'
```
50Wliu: Why was this changed?

MaximSokolov: `\w` is enough here, as I modified `end` pattern to allow regular functions/properties.

CentricStorm:
> Looks legit to me


Yeah I should have been clearer - the comments in that example image were only referring to the lines immediately after them - `console//` is working fine :thumbsup:.

Unfortunately there is still a problem though:

untitled

MaximSokolov:
>
```js
console . loga()
```

Fixed, thanks!

CentricStorm:
More issues:

untitled

50Wliu:
Looks good to me except for one thing.  Adding a comment before the period does not cause the comment to get recognized, eg `console/*hi*/.log()`.  This was evidently working according to @CentricStorm's previous screenshot, but now it isn't.

Some things that we might want to address in future PRs though:
* `console` and `Math`-specific methods aren't being tokenized with `entity.name.function.js`.
* Adding a comment after the period causes the match to end prematurely, eg `console./*hi*/log()`.

CentricStorm:
> Looks good to me except for one thing. Adding a comment before the period does not cause the comment to get recognized, eg `console/*hi*/.log()`. This was evidently working according to @CentricStorm's previous screenshot, but now it isn't.

Working fine for me on Atom 1.8.0:

untitled

> - `console` and `Math`-specific methods aren't being tokenized with `entity.name.function.js`.

What should they be scoped as? With the `entity.name.function.js` scope, they will be indistinguishable from normal function calls. Some colour schemes apply styling to `.entity.name.function`, which is higher precedence than `.support.function`, so will override.

> - Adding a comment after the period causes the match to end prematurely, eg `console./*hi*/log()`.

Some things are impractical to do with regular expressions. If 100% syntax highlighting accuracy is desired then it would be easier to change to a language-specific parser.

MaximSokolov:
> Adding a comment before the period does not cause the comment to get recognized, eg
```js
console/*hi*/.log()
```

Are you sure you've tested this branch? It should be working in both f625823 and 7bddd57
Now `.log()` is also scoped properly (`support.function`)

MaximSokolov:
Agree with @CentricStorm about scopes and `console./*hi*/log()` case

50Wliu:
I'm pretty sure I was...I specifically checked out the branch on Github Desktop :confused:.  I'll give it another go today.

PR #362, opened 2016-04-15
Highlight GraphQL inside Relay.QL template strings

charliesome:
This pull request adds a rule to highlight GraphQL inside `Relay.QL` template strings. 

As of https://github.com/github/linguist/pull/2947 we have a GraphQL grammar in Linguist.

cc @kevinsawicki @arfon @kdaigle @gjtorikian

### DIFF
--- a/grammars/javascript.cson
+++ b/grammars/javascript.cson
@@ -1218,6 +1218,31 @@
         ]
       }
       {
+        'begin': '(Relay\\.QL)\\s*(`)'
+        'beginCaptures':
+          '1':
+            'name': 'entity.name.function.js'
+          '2':
+            'name': 'punctuation.definition.string.begin.js'
+        'end': '`'
+        'endCaptures':
+          '0':
+            'name': 'punctuation.definition.string.end.js'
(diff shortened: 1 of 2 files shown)


grammars/javascript.cson
```diff
@@ -1218,6 +1218,31 @@
         ]
       }
       {
+        'begin': '(Relay\\.QL)\\s*(`)'
```
kevinsawicki: Really minor, but should `\\.` be captured as `'meta.delimiter.method.period.js'` so it it styled similarly to method invocations?

charliesome: Good call, let me fix that right up

charliesome: Upon further discussion we've decided to not do this for now, as the entire `Relay.QL` is a marker saying that the contents of the template string is GraphQL source. We can always capture the `.` separately later on if we decide we want to.

kevinsawicki:
:+1: Thanks for adding this :ship:

gjtorikian:
Nice!

PR #365, opened 2016-04-17
Don't scope decimal mark to allow ligatures to work

### LINKED ISSUE #364: OpenType features on decimal separators are disabled
This is how a number renders:
```html
<span class="constant numeric decimal js">
  42<span class="meta delimiter decimal period js">.</span>2
</span>
```

Because the decimal is in a separate element, OpenType contextual replacement feature does not work across the entire number. For instance, a typeface might contain two different period glyphs – one for text, and the other for decimal separator.

This is how the DOM needs to look for the feature to work:
```html
<span class="constant numeric decimal js">
  42.2
</span>
```

I understand that somebody might be relying on this extra element for, perhaps, syntax highlighting the decimal separator. There seems to be a clear conflict of features, and it is up to you to make the call.

MaximSokolov:
Closes #364


50Wliu:
I'm :-1: on this.  I personally prefer my decimals highlighted differently to let them stand out.

JonathanWolfe:
I'm also against this as neither Hasklig or FiraCode have ligatures that operate on decimals. This would only be removing highlighting without gain.

PR #368, opened 2016-04-19
Prioritize HTML comments over regular comments

50Wliu:
`-->` automatically ends a regular `//` comment for HTML interoperability.  For example,
```html
<script type="text/javascript"><!--
       var hello;
//--></script>
```
is valid HTML and JavaScript and ends the `<script>` block correctly.

This does **not** add support for [ES6 HTML-style comments](http://www.ecma-international.org/ecma-262/6.0/#sec-html-like-comments).

Unblocks atom/language-html#90.

### DIFF
--- a/grammars/javascript.cson
+++ b/grammars/javascript.cson
@@ -767,15 +767,6 @@
     'include': '#comments'
   }
   {
-    'captures':
-      '0':
-        'name': 'punctuation.definition.comment.html.js'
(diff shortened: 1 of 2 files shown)


50Wliu:
Misunderstood the intended behavior.  `//` automatically ends upon reaching a `</script>` tag.  This makes the problem much harder to solve though.

PR #371, opened 2016-04-27
Add support for ES6 HTML-style comments

50Wliu:
Also end comments when a `</script>` tag is encountered.  This is because HTML and JavaScript can intelligently detect when to break out of the comment and end the script block.  Unfortunately, since we're using a 100% regex-based parser, there's really not much we can do to detect if we're embedded (writing this just now gave me an idea though...will look into it later).  So for now we just be as conservative as possible.

Unblocks atom/language-html#90


JonathanWolfe:
Do you need an `Atom Member` to review or am I allowed to?

50Wliu:
Of course you can review it.  Though I'm pretty sure I was dissatisfied with this approach and/or something was wrong with it.  Should probably have documented that somewhere because now I can't remember.

JonathanWolfe:
Well if you still want to bring this in, just rebase the branch and I'll check the spec and make some comments on improvements.

PR #375, opened 2016-05-20
Replace 'function' with Arrows

PatrickHeneise:


JonathanWolfe:
Wouldn't it be better to create additional snippets for arrow function instead of replacing the regular function ones?

thomasjo:
> Wouldn't it be better to create additional snippets for arrow function instead of replacing the regular function ones?

Either that or provide alternate snippets for the plain function, so those who aren't running on Node (and/or transpiling with Babel or similar) aren't left out.

I'm all for moving to the future, but we have to realize there are still a lot of developers who simply cannot go there yet.

JonathanWolfe:
Not to mention that arrow function and regular functions are fundamentally different and are not always interchangeable.

PR #378, opened 2016-06-15
Add @public docblock match in Javascript grammar

ivangabriele:
Since it exists in official JSDoc dictionary :

* http://usejsdoc.org/tags-public.html


50Wliu:
Thanks!

PR #393, opened 2016-08-04
🐛  Fix 'constructor' boundary

### LINKED ISSUE #391: wrong highlighting for words that starts with "constructor"
currently at version 1.9.1

if a key starts with "constructor", only "constructor" will be highlighted as different color, also screw up highlighting for other keys in the object

Using the Base16 Tomorrow Dark theme (bug is independent of theme)
```js
a = {
  constructorExtension: {  // result will be <blue>constructor</blue>Extension
    keys: {},  // keys will be white
    'morekeys': {},  // 'morekeys' will be white when it is usually green
    someFn: function(){},  // someFn is usually blue, but is now white
    'otherFn': function(){},  // the quotes are supposed to be green and otherFn blue, but is all white
  },
};
```
compare against
```js
b = {
  construXXXXctorExtension: {  // bug do not occur if construXXXXctor
    keys: {},  
    'morekeys': {},  
    someFn: function(){},  
    'otherFn': function(){}, 
  },
};
```

MaximSokolov:
```js
a = {
  constructorABC: {}
};
```

Fixes #391

### DIFF
--- a/grammars/javascript.cson
+++ b/grammars/javascript.cson
@@ -416,7 +416,7 @@
     ]
   }
   {
-    'begin': '(?=\\b(constructor)\\s*)'
+    'begin': '(?=\\bconstructor\\b\\s*)'
     'end': '(?<=})'
     'patterns': [
       {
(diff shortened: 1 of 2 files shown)


50Wliu:
I think this deserves some specs.

MaximSokolov:
Done: 883e1e0

PR #396, opened 2016-08-09
Add support for JSDoc parameter properties

bgriffith:
This PR adds support for JSDoc's parameter properties:

`@param {type} parameter.property description`


50Wliu:
I think you want `\\.`, not just `.`.

bgriffith:
@50Wliu Indeed you're right! Not sure how I let that slip through :+1:

PR #401, opened 2016-08-10
Add JSDoc support for optional parameter types

bgriffith:
This PR adds support for JSDoc highlighting of optional parameter types, defined by suffixing the type with `=` (http://usejsdoc.org/tags-type.html)

`@param {type=} variable - description`

After:

screen shot 2016-08-10 at 20 06 44


50Wliu:
The `=` does not need to be escaped.

bgriffith:
@50Wliu Updated!

50Wliu:
Upon looking at it again, `=?` should suffice, right?  No need for a non-capturing group?

bgriffith:
Heh indeed! Trimmed it down.

50Wliu:
:+1:

PR #402, opened 2016-08-11
Fixed autocompletion error in "console."

Jamozzles:
Autocompleting "console." would result in "console.console."


50Wliu:
This is intended as they are _snippets_, not autocompletions.  You're supposed to type the snippet (for example, `log`), and then press tab or select it through autocomplete-plus to trigger the snippet.

PR #404, opened 2016-08-15
Extend list of DOM methods

Zirro:
This pull request extends the list of methods recognized as being part of the DOM. The original list was very incomplete and led to peculiar behaviour such as classList.add() and classList.remove() being coloured differently.

These additions were collected by iterating through several common DOM interfaces, mainly in Firefox and Chrome. For practical reasons it is by no means complete, but I believe it covers those which are most-used. Intentional omissions include prefixed properties and non-standard properties with little adoption. I also removed a couple of overlapping properties from the support.function.js regex.

### DIFF
--- a/grammars/javascript.cson
+++ b/grammars/javascript.cson
@@ -1442,7 +1442,7 @@
                 'match': '''(?x)
                   \\b(acceptNode|add|addEventListener|addTextTrack|adoptNode|after|animate|append|
                   appendChild|appendData|before|blur|canPlayType|captureStream|
-                  caretPositionFromPoint|caretRangeFromPoint|checkValidity|clear|clear|click|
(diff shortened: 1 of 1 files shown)


50Wliu:
For stuff like `(exec|execScript)`, that won't work and will exit as soon as it reaches `exec`.  So you have two options:
1. `(exec(script)?)`
2. `(execScript|exec)`

Zirro:
While I moved execCommand() in this PR, I don't think I changed the (exec|execScript)-logic. Furthermore, .exec(), .execScript() and .execCommand() all seem to highlight and receive the correct classes when I test in Atom.

50Wliu:
Hmm, you're right - sorry about that.  Just waiting for CI at this point.

50Wliu:
Don't know what's up with Travis right now -.-

PR #405, opened 2016-08-19
Fix missing highlighting in case statements

Alhadis:
Some values don't receive their usual highlighting when used inside a case statement:

Figure 1

Not all of these statements are valid JavaScript (e.g., `case debugger:`), but authors should probably be reminded of a keyword's significance through highlighting.

### DIFF
--- a/grammars/javascript.cson
+++ b/grammars/javascript.cson
@@ -275,11 +275,19 @@
     'name': 'meta.export.js'
   }
   {
-    'match': '(?<!\\.)\\b(super|this|arguments)(?!\\s*:)\\b|(?<=\\?)(?:\\s*)(super|this|arguments)(?=\\s*:)'
+    'match': '(?x)
+      (?<!\\.)\\b(super|this|arguments)(?!\\s*:)\\b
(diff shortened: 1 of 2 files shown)


50Wliu:
Can you triple-quote the extended regexes?

Alhadis:
Ah, sorry! Fixed. =)

You might wanna fix the expressions on lines [53](https://github.com/Alhadis/language-javascript/blob/f300a4de2eacda4ce671adbb1c041da7fcde8129/grammars/javascript.cson#L53), [89](https://github.com/Alhadis/language-javascript/blob/f300a4de2eacda4ce671adbb1c041da7fcde8129/grammars/javascript.cson#L89), and a few others too...

Alhadis:
(I would've changed 'em, but I didn't want this PR changing unrelated code...)

50Wliu:
When I have time™ I guess :P.

Alhadis:
[Take some of mine](https://github.com/atom/language-javascript/pull/407), I've got plenty.

PR #408, opened 2016-08-20
Fix arrow functions inside ternary expressions

### LINKED ISSUE #279: no highlight for `case null:` inside `switch` block
`null` is highlighted in most common cases except inside of a switch-case block

image

### LINKED ISSUE #406: Arrow functions at the end of ternary expressions are not recognized
Using Atom 1.9.9

Example code:

```js
var result = condition ? something : (arg1, arg2) => new Foo(arg1, arg2);
```

Atom renders it like this:

Note that the arrow function body is not highlighted at all (e.g. `new` should be identified as a keyword).

This issue also occurs when using curly brackets for an expression body, so the following code also displays the bug:

```js
var result = condition ? something : (arg1, arg2) => { return new Foo(arg1, arg2) };
```

However, it seems to be fixed if the curly brackets are followed by a newline, so the following code is highlighted correctly:

```js
var result = condition ? something : (arg1, arg2) => {
  return new Foo(arg1, arg2)
};
```

MaximSokolov:
Fixes #406 by using begin/end matching for ternary expressions
```js
result = condition ? something : (a, b) => a + b
```

Fixes #279 by matching case statements only in switch block (replaces #405 fix)
```js
switch (foo())
{
    case 'string':
    case a:
    case a + b:
    case foo(bar):
    case ABC:
    case 123:
      foo(bar)
      break;
    case null:
    case true:
    default:
}
case abc: // won't be highlighted
```

### DIFF
--- a/grammars/javascript.cson
+++ b/grammars/javascript.cson
@@ -279,15 +279,11 @@
       (?<!\\.)\\b(super|this|arguments)(?!\\s*:)\\b
       |
       (?<=\\?)\\s*(super|this|arguments)(?=\\s*:)
-      |
-      (?<=[\\s}:;]case|^case)\\s+(super|this|arguments)(?=\\s*:)
     '''
     'captures':
       '1':
         'name': 'variable.language.js'
       '2':
-        'name': 'variable.language.js'
-      '3':
         'name': 'variable.language.js'
   }
   {
@@ -899,12 +895,15 @@
     'name': 'meta.control.yield.js'
   }
   {
-    'match': '(?<!\\.)\\b(await|break|case|catch|continue|do|else|finally|for|if|import|package|return|switch|throw|try|while|with)(?!\\s*:)\\b'
+    'match': '(?<!\\.)\\b(await|break|catch|continue|do|else|finally|for|if|import|package|return|throw|try|while|with)(?!\\s*:)\\b'
     'name': 'keyword.control.js'
   }
   {
     'match': '(?<!\\.)\\b(default)\\b'
     'name': 'keyword.control.js'
+  }
+  {
+    'include': '#switch_statement'
   }
   {
     'match': '(?<!\\.)\\b(delete|in|of|instanceof|new|typeof|void)(?!\\s*:)\\b'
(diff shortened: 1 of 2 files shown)


Alhadis:
Uhm, what was wrong with my fix, exactly?

Alhadis:
There's no need to restrict #279's fix to `switch` blocks, because `case` statements won't appear anywhere else. I feel your approach is over-complicating things.

MaximSokolov:
This is a necessary step to recognize object keys. See #307
```js
// for e.g.
switch() {
  default:
}
a = {
  default: 123
}
```

Alhadis:
Sigh... I see this was already discussed.

In which case, I'll probably have to rebase against your branch, since I'm in the middle of fixing numerous bugs for another PR...

Alhadis:
@50Wliu Any chance of getting this merged soon? There'll be a [new version of Linguist](https://github.com/github/linguist/pull/2773#issuecomment-241062154) this week, and it'd be nice to have these highlighting bugs fixed on the site. Otherwise, they'll linger on GitHub for a few weeks longer. :p

Just to be clear: once this and the other PR get merged, I'll go balls out on remaining bug fixes in the hopes of getting everything under the one release. =)

/cc @pchaigno Paging you because @arfon won't hear me from where I'm standing. Literally (his name's not showing up in my mentions menu...).

50Wliu:
> Any chance of getting this merged soon?

I am currently much busier than I was just a week ago, so expect my activity on GitHub to tank significantly for a while.  I'll try to merge this to unblock your other PRs but those might sit for some time before proper review.

grammars/javascript.cson
```diff
@@ -1023,6 +1004,29 @@
...
+    'beginCaptures':
+      '0':
+        'name': 'keyword.operator.ternary.js'
+    'end': ':'
+    'endCaptures':
+      '0':
+        'name': 'keyword.operator.ternary.js'
+    'patterns': [
+      {
+        'match': '(\\w+)(?=\\s*:)'
```
50Wliu: Can you explain what this match does in more detail?

In addition, why is the second `$self` needed?

MaximSokolov: See 157241d0fd3c8. There is no need to include `$self` here. Though most of the rules must be moved to the repo and it also requires addition testing, so I kept it as is for now. The second `$self` includes everything else between `?` and `:`

50Wliu:
7296d5d yesssss

Glad to see you're working on this again @MaximSokolov!

PR #418, opened 2016-08-23
Allow multiple types to be specified in JSDoc tags

Alhadis:
JSDoc [allows](http://usejsdoc.org/tags-type.html#jsdoc-types) more than one type to be specified in a `@type` tag or whatever:

Figure 1

This PR fixes that.


50Wliu:
The documentation also says that the types have to be surrounded by parentheses:
`{(number|boolean)}`

Alhadis:
I'll add support for the bracket-notation, but I should mention the notation doesn't appear to be mandatory. Running `jsdoc` on this produces the same output:

```js
/**
 * Insert snarky name here.
 *
 * Add dry-witted description here.
 *
 * @param {Number|Function} callable - Something to return
 * @return {String|Object} Something
 */
function test(callable){
	return "ffs";
}
```

Result:

See?

Zirro:
There was a brief discussion about this here: https://github.com/jsdoc3/jsdoc/issues/1104

The TL;DR-version is that both ways are accepted.

Alhadis:
Okay, fixed.

grammars/javascript.cson
```diff
@@ -1701,7 +1701,7 @@
         'match': '''(?x)
           (?:(?<=@returns)|(?<=@return))
           \\s+
-          ({(?:\\*|(?:\\?|\\!|\\.{3})?[a-zA-Z_$][\\w$]*(?:\\.[a-zA-Z_$][\\w$]*)*(?:\\[\\])?=?)})
+          ({(?:\\*|(?:\\?|\\!|\\.{3})?[a-zA-Z_$][\\w$]*(?:[\\.|][a-zA-Z_$][\\w$]*)*(?:\\[\\])?=?)})
```
50Wliu: What about parentheses here?

Alhadis: Fixed.

PR #419, opened 2016-08-24
JSDoc fixes - Modify defaults, param name optional and new type support

bgriffith:
Few fixes in this PR :tada:

## Fix 1 - Modify defaults
This PR switches what JSDoc defaults to when deciding how many captures to check for. Currently it defaults to 4 with one requiring an `@return` or `@returns` tag to only check for 3.

Having read the docs further, it's become obvious that 3 should be the default for all tags except for when we have a `@param` or `@type` which then requires 4.

## Fix 2 - Make param name optional

I've also made the param name optional and this change fixes an issue with tags that only have a type specified.

Before:
screen shot 2016-08-24 at 21 29 00

After:
screen shot 2016-08-24 at 21 29 29

## Fix 3 - Add support for types with a set of type arguments

Adds support a type with a set of type arguments. I.e. `Array<string>` an array of strings.

Before:
screen shot 2016-08-24 at 22 54 50

After:
screen shot 2016-08-24 at 22 50 52

## Fix 4 - Add support for unknown type

Adds support for the unknown type `{?}`

Before:
screen shot 2016-08-24 at 22 59 58

After:
screen shot 2016-08-24 at 23 03 12

## Fix 5 - Add support for key-value pair types

Adds support for specifying the types of key-value pairs within the parameter.

Before:
screen shot 2016-08-24 at 23 25 46

After:
screen shot 2016-08-24 at 23 25 26

## Lastly 
:art: Added whitespace to break up extremely long JSDoc rules and increase readability. (Please let me know if there's a specific style I should follow!)


50Wliu:
Can you also add some comments for people not familiar with the JSDoc syntax?  For example, adding a comment after the `\\?` to signify that that's the unknown type.

grammars/javascript.cson
```diff
@@ -1781,9 +1781,37 @@
...
       {
         'match': '''(?x)
-          ({(?:\\*|(?:\\?|\\!|\\.{3})?(?:\\([a-zA-Z_$][\\w$]*(?:[\\.|][a-zA-Z_$][\\w$]*)*\\)|[a-zA-Z_$][\\w$]*(?:[\\.|][a-zA-Z_$][\\w$]*)*)(?:\\[\\])?=?)})
+          (?:(?<=@param)|(?<=@type))
           \\s+
-          (\\[(?:[a-zA-Z_$]+(?:=[\\w][\\s\\w$]*)?)\\]|(?:[a-zA-Z_$][\\w$]*(?:[\\.|][a-zA-Z_$][\\w$]*)*))
+          ({(?:
+            \\* |
+            \\? |
+            (?:\\?|\\!|\\.{3})?
```
50Wliu: I don't think the ! needs to be escaped.

grammars/javascript.cson
```diff
@@ -1781,9 +1781,37 @@
...
+            (?:
+              \\(
+                [a-zA-Z_$]+
+                (?:[\\w$]*|\\<[\\w$]+(?:\\,\\s+[\\w$]+)*\\>)
```
50Wliu: `<`, `>`, and `,` also don't need to be escaped

bgriffith:
@50Wliu Hey! Sure will do. Can you give me a heads up on how to comment within an extended match? What's the syntax? I can't seem to find an example.

Alhadis:
`(?# This)` or `# This`.

Also, [install this](https://atom.io/packages/regex-comments).

50Wliu:
Just `<your regex> # comment` in an extended regex.  As long as the comment is at the end of the line you're fine.

bgriffith:
@Alhadis @50Wliu Cheers guys :+1:

bgriffith:
@50Wliu  Something like that ok?

50Wliu:
Yeah - that helps so much!

PR #423, opened 2016-08-31
Add \t \r \n \v & \f to RegEx character-class scope

MoritzKn:
These are also character classes:
- \t (horizontal tab)
- \r (carriage return)
- \n (linefeed)
- \v (vertical tab)
- \f (form-feed)

See: [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp#Special_characters_meaning_in_regular_expressions)


50Wliu:
Thanks!

PR #424, opened 2016-08-31
Improve JSDoc support for optional and default values

bgriffith:
This PR fixes 2 issues with optional and default values when using the param square bracket syntax and when using valid whitespace.

```js
/**
 * @param {string} [foo] - optional value
 * @param {string} [foo=bar] - default value
 */
```

### Before
screen shot 2016-08-31 at 23 32 40

### After
screen shot 2016-08-31 at 23 31 55

### Before
screen shot 2016-08-31 at 23 34 49

### After
screen shot 2016-08-31 at 23 35 14

### DIFF
--- a/grammars/javascript.cson
+++ b/grammars/javascript.cson
@@ -1830,6 +1830,7 @@
           \\s+
           (
             \\[                                          # [foo] optional parameter
+              \\s*
               (?:
                 [a-zA-Z_$][\\w$]*
                 (?:
@@ -1837,10 +1838,12 @@
                   [a-zA-Z_$][\\w$]*
                 )*
                 (?:
+                  \\s*
                   =                                      # [foo=bar] Default parameter value
                   [\\s\\w$]*
                 )?
               )
+              \\s*
             \\] |
             (?:
               [a-zA-Z_$][\\w$]*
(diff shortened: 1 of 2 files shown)


grammars/javascript.cson
```diff
@@ -1830,7 +1830,20 @@
...
+              (?:
+                [a-zA-Z_$][\\w$]*
+                (?:
+                  \\.                                    # Foo.Bar namespaced parameter
+                  [a-zA-Z_$][\\w$]*
+                )*
+                (?:
+                  \\s*
+                  =                                      # [foo=bar] Default parameter value
+                  [\\s\\w$]*
```
50Wliu: I think this should be `\\s*=\\s*[\\w$]*` instead.

bgriffith: Only issue with this is that it won't allow spaces within the default value;

```
[bar=baz qux]
```

bgriffith: I could do `\\s*=\\s*[\\w$\\s]*` if it helps readability.

50Wliu: Yeah, I think that would help.

bgriffith:
@50Wliu Updated :+1:

50Wliu:
Thanks!

PR #428, opened 2016-09-02
Allow optional prefix with JSDoc's type application

bgriffith:
This PR allows one to use an optional `.` prefix when defining type application with JSDoc.

So we now support: `{Array.<string>}` as well as `{Array<string>}` ((http://usejsdoc.org/tags-type.html)
### Before

screen shot 2016-09-02 at 21 23 33

### After

screen shot 2016-09-02 at 21 22 51

### DIFF
--- a/grammars/javascript.cson
+++ b/grammars/javascript.cson
@@ -1801,7 +1801,7 @@
                     [\\w$]*
                     (?:\\[\\])?                          # {(string[]|number)} type application, an array of strings or a number
                   ) |
-                  <[\\w$]+(?:,\\s+[\\w$]+)*>             # {Array<string>} or {Object<string, number>} type application
+                  \\.?<[\\w$]+(?:,\\s+[\\w$]+)*>         # {Array<string>} or {Object<string, number>} type application (optional .)
                 )
                 (?:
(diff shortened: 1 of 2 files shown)


50Wliu:
Where is this documented?  I looked at the page you provided but couldn't find anything about the `.` being optional.

bgriffith:
Sorry for the delay! Took a few weeks off 😃 

It's found in the google closure compiler type expressions which JSDoc supports - https://github.com/google/closure-compiler/wiki/Annotating-JavaScript-for-the-Closure-Compiler#type-application

grammars/javascript.cson
```diff
@@ -1821,14 +1821,14 @@
                   [\\w$]*
                   (?:\\[\\])?                            # {string[]|number} type application, an array of strings or a number
                 ) |
-                <[\\w$]+(?:,\\s+[\\w$]+)*>               # {Array<string>} or {Object<string, number>} type application
+                \\.?<[\\w$]+(?:,\\s+[\\w$]+)*>           # {Array.<string>} or {Object.<string, number>} type application (optional .)
```
50Wliu: Inconsistent comment compared to the others

bgriffith: Comment updated to be consistent.

PR #436, opened 2016-09-29
added .ts to grammers/javascript.cson for TypeScript language support

siunegu:
Not sure if this will work, but just wanted to add potential TypeScript support for the syntax highlighter.

### DIFF
--- a/grammars/javascript.cson
+++ b/grammars/javascript.cson
@@ -13,6 +13,7 @@
   'jspre'
   'pac'
   'pjs'
+  'ts'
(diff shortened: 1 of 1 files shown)


50Wliu:
This won't work as TypeScript is a superset of JavaScript.  Hence, applying Javascript highlighting to Typescript files won't end up that well.

I suggest taking a look at https://atom.io/packages/atom-typescript or https://atom.io/packages/language-typescript-grammars-only instead.

PR #438, opened 2016-10-06
Return control to language-html when </script> is detected when embedded

50Wliu:
When embedded, a </script> tag will break out of the script block even if it would otherwise be in a comment or string.  There's no way to catch when we're embedded with javascript.cson, but we can do it using an injectionSelector.

Supersedes and closes #371.
Blocked on atom/first-mate#78.
Unblocks atom/language-html#90.


50Wliu:
No tests are included for this change because they belong in language-html.

50Wliu:
This is actually causing first-mate to throw some errors because this is technically a zero-width match, so it'll match over and over again.  I'll get this right one day...

PR #447, opened 2016-10-17
Improve support for property highlighting

Alhadis:
__Adds:__  

*	Highlighting to `get`/`set` operators:

	```js
	set name(i){ }
	get name( ){ }
	```

*	Support for computed method names (ES6):

	```js
	const object = {
		    [ "na" + "me" ](){ }
		get [ "na" + "me" ](){ }
	};
	```

__Preview:__  
Before and after

__Question:__
How long until the scope-names of `[`, `{` and `(` can be fixed to use `punctuation` instead of the painfully-incorrect `meta.brace`...? It's inconsistent not just with other grammars, but also in how it affects highlighting between syntax themes.

I mean, this says it all:

Pls


50Wliu:
New scopes use `punctuation`, old scopes in patterns that are updated are generally updated to use `punctuation` as well.  So in this case you can change them to punctuation.

grammars/javascript.cson
```diff
@@ -468,10 +468,25 @@
...
+          \\]
+        )
+        \\s*\\(\\s*               # Start of arguments list
+          (
+            "[^"]*"  |            # Double-quoted string
+            '[^']*'  |            # Single-quoted string
+             [^"()']              # Any non-bracket or non-quote
+          )*
+        \\)\\s*                   # End of arguments
+        \\{                       # Beginning of body
```
50Wliu: {

Alhadis:
Ace. Give me a sec.

grammars/javascript.cson
```diff
@@ -488,6 +503,29 @@
         'name': 'meta.function.method.definition.js'
         'patterns': [
           {
+            'match': '(\\[)(.+)(\\])(?=\\s*\\()'
+            'captures':
+              '1':
+                'name': 'meta.brace.square.js'
```
50Wliu: Use punctuation scopes

grammars/javascript.cson
```diff
@@ -488,6 +503,29 @@
...
+                  {
+                    'include': '$self'
+                  }
+                  {
+                    'match': '[a-zA-Z_$][\\w$]*'
+                    'name': 'variable.parameter.property.js'
+                  }
+                ]
+              '3':
+                'name': 'meta.brace.square.js'
```
50Wliu: Use punctuation scopes

grammars/javascript.cson
Alhadis: Heh, I always forget to unescape those for some reason...

50Wliu:
Can there also be a test with just a computed property name (without a getter/setter in front)?

Alhadis:
Done. Really thankful I [wrote this](https://github.com/Alhadis/Atom-PhoenixTheme/commit/85f793feb8ed93cc4fb67b9110e8dd7266bb2bc6), BTW. :p

50Wliu:
> Done. Really thankful I wrote this, BTW. :p

To each their own :).  Just waiting for the build at this point.

PR #450, opened 2016-10-23
Missing single line comment <!--

brunoduran:
There are two ways of using single line comments in Javascript:

1. The usual way with two consecutive slashes `//`
2. An obscure but also valid `<!--` as the start of the html block comments

The following two lines are completely valid in Javascript:

```javascript
1 + 1 == 2 // usual
1 + 1 == 0 <!-- obscure
```

### DIFF
--- a/grammars/javascript.cson
+++ b/grammars/javascript.cson
@@ -2086,7 +2086,7 @@
         'end': '(?!\\G)'
         'patterns': [
           {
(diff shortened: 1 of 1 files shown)


50Wliu:
This will be superseded by #438 so I'm going to close this one in favor of that one, which should be merged shortly.

PR #453, opened 2016-10-29
Fix #452: Wrong scopes for methods named "import"

### LINKED ISSUE #452: `import` can be used as a method name, and this confuses the grammar.
Version: 0.119.0

If `import` is used as a method name, the grammar gets confused. For example, take this file:

```javascript
class Blort {
  normal(x) {
    if (x > 10) {
      error(`${x} is too big!`);
    }    
  }

  import(x) {
    if (x > 10) {
      error(`${x} is too big!`);
    }
  }
}
```

Here's how it gets rendered for me:

rendered

You can see that the first method gets proper markup but the second doesn't. In this case, it seems to miss that first backtick as an open-quote-ish thing, so the rest of the file after the second backtick is taken to be an unterminated string.

Alhadis:
Fixes #452:

```js
import(x) { "ABCDEFGHIJKLMNOPQRSTUVWXYZ"; }
Import(x) { "ABCDEFGHIJKLMNOPQRSTUVWXYZ"; }
```

### DIFF
--- a/spec/javascript-spec.coffee
+++ b/spec/javascript-spec.coffee
@@ -1249,6 +1249,14 @@ describe "Javascript grammar", ->
       {tokens} = grammar.tokenizeLine('"func" : a => a')
       expect(tokens[8]).toEqual value: '=>', scopes: ['source.js', 'meta.function.arrow.json.js', 'storage.type.function.arrow.js']
 
+      {tokens} = grammar.tokenizeLine('import (x) { return "Yeah"; }')
+      expect(tokens[0]).toEqual value: 'import', scopes: ['source.js', 'meta.function.method.definition.js', 'entity.name.function.js']
+      expect(tokens[11]).toEqual value: 'Yeah', scopes: ['source.js', 'string.quoted.double.js']
+
+      {tokens} = grammar.tokenizeLine('export (x) { return "Nah"; }')
+      expect(tokens[0]).toEqual value: 'export', scopes: ['source.js', 'meta.function.method.definition.js', 'entity.name.function.js']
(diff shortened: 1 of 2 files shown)


spec/javascript-spec.coffee
```diff
@@ -1249,6 +1249,14 @@ describe "Javascript grammar", ->
...
       expect(tokens[8]).toEqual value: '=>', scopes: ['source.js', 'meta.function.arrow.json.js', 'storage.type.function.arrow.js']
 
+      {tokens} = grammar.tokenizeLine('import (x) { return "Yeah"; }')
+      expect(tokens[0]).toEqual value: 'import', scopes: ['source.js', 'meta.function.method.definition.js', 'entity.name.function.js']
+      expect(tokens[11]).toEqual value: 'Yeah', scopes: ['source.js', 'string.quoted.double.js']
+
+      {tokens} = grammar.tokenizeLine('export (x) { return "Nah"; }')
+      expect(tokens[0]).toEqual value: 'export', scopes: ['source.js', 'meta.function.method.definition.js', 'entity.name.function.js']
+      expect(tokens[11]).toEqual value: 'Nah', scopes: ['source.js', 'string.quoted.double.js']
+
```
MaximSokolov: These two tests should be in `tokenizes ES6 method definition` section

Alhadis:
... [uh, what the hell](https://s3.amazonaws.com/archive.travis-ci.org/jobs/171638610/log.txt)?

> No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
The build has been terminated

PR #464, opened 2016-11-23
Delete redundant .coffeelintignore file

Alhadis:
Added in 2058cd0, but there isn't a `spec/fixtures` folder (and to my understanding, there never was). This file is just adding clutter.


mnquintana:
Nice catch! Thanks for cleaning this up 👍

PR #465, opened 2016-11-27
Revert "Return control to language-html when </script> is detected when embedded"

50Wliu:
Reverts atom/language-javascript#438
See last comment there for justification.


Alhadis:
To my current understanding, the best way to do this is the way I did it in language-xml#54 - create a dummy match for everything leading up to a break-off point (`</script>`), and embed the contents of a `match` rule using the embedded grammar itself.

PR #466, opened 2016-11-29
Update javascript.cson to style template strings prefixed with a 'css' tag

trevorhreed:
Currently, when you prefix a template literal with an `html` tag, you get HTML syntax highlighting in that template literal. This doesn't work for CSS. This pull request adds support for CSS syntax highlighting in template literals prefixed with `css`, `CSS`, or `Css`.

Before:
screen shot 2016-11-28 at 9 29 37 pm

After:
screen shot 2016-11-28 at 9 29 48 pm

### DIFF
--- a/spec/javascript-spec.coffee
+++ b/spec/javascript-spec.coffee
@@ -17,6 +17,8 @@ describe "JavaScript grammar", ->
 
     runs ->
       grammar = atom.grammars.grammarForScopeName("source.js")
+      console.log typeof grammer
+      console.dir grammer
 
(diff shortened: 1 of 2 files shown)


spec/javascript-spec.coffee
```diff
@@ -17,6 +17,8 @@ describe "JavaScript grammar", ->
 
     runs ->
       grammar = atom.grammars.grammarForScopeName("source.js")
+      console.log typeof grammer
```
Alhadis: Delete this.

PR #463, opened 2016-11-23
Fix #460: Incorrect highlighting for multiline imports

Alhadis:
Fixes the issues brought up by @bgriffith in #460, as well as another I noticed while filling out specs:

```js
import { default as alias };

import
{
default
as
alias
};
```

PR #468, opened 2016-12-04
Fix quoted defaults / inline links in JSDoc @param lines

Alhadis:
Two JSDoc-related fixes, both of which are legal syntax:

```js
/**
 * @param {String} [name="default value"] Description
 * @param {Object} name - Description with {@link link}.
 */
```

**Preview:**

Spiffy, innitt?


50Wliu:
Needs a test for single quotes.  Otherwise looks good.

Alhadis:
Done. You can merge #463 already too, no idea why that's still sitting open.

PR #469, opened 2016-12-04
Allow OpenType ligatures in "double-not" operators

Alhadis:
This is such a common trick in JavaScript, and it looks pretty nice as a ligature:

A patched typeface currently can't target `!!` with a ligature, because each `!` is tokenised separately. The scopes are the same, so there's otherwise no outward difference. For that reason, I've not bothered to include a spec; please advise if you'd like one.


50Wliu:
While technically not an operator, I guess I'm ok with this.

PR #477, opened 2016-12-31
Add ternary operator snippet

AlexKvazos:
Added a ternary operator snippet that can be accessed with `tern`. This are strongly used in react to control the rendering of two components based on a simple condition.

### DIFF
--- a/snippets/language-javascript.cson
+++ b/snippets/language-javascript.cson
@@ -11,6 +11,9 @@
   'do':
     'prefix': 'do'
     'body': 'do {\n\t$2\n} while (${1:true});'
+  'condition ? true : false':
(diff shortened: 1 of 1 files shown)


50Wliu:
Looks like the snippets are in alphabetical order by prefix: can you move the new snippet?

And also, what do you think about a simpler `${1:condition} ? ${2:true} : ${3:false}`?  Parentheses can be added as needed.

AlexKvazos:
I removed the parentheses, though I do not see an alphabetical order being followed.

50Wliu:
And that is what I get for only looking at a portion of the file.  Thanks for this!

Issue #471, opened 2016-12-12
`const` has a different class from `var` and `let`


kaicataldo:
`const` currently has the `modifier` class as opposed to the `type` class that `var` and `let` do, leading to `const` being highlighted differently. Is this intentional? It would make a lot more sense to me if they had the same classes and were styled the same.

<img width="849" alt="screen shot 2016-12-11 at 8 49 13 pm" src="https://cloud.githubusercontent.com/assets/7041728/21085758/7aed541c-bfe3-11e6-97db-53c57536eecc.png">
<img width="793" alt="screen shot 2016-12-11 at 8 49 45 pm" src="https://cloud.githubusercontent.com/assets/7041728/21085759/7aee1d0c-bfe3-11e6-8a8a-f265319b8fba.png">

Happy to look into making a PR for this, if the team is amenable to this change.

Alhadis:
Speaking as a grammar developer, I can attest that `storage.modifier` is a semantically inaccurate choice of scope, too.

It's supposed to be used for tokens like `public` or `protected` which *modify* how a value is stored/represented by the language.

So `const` should actually have the same scopes, because it fills the same lexical role as `let` and `var` do - named data assignment.

50Wliu:
This was probably copied over from other languages, where const is actually a modifier (like in Java).  So yeah, I'm ok with this change.

Alhadis:
Are you implying JavaScript ever borrowed anything from Java?

COME ON SON.

PR #479, opened 2017-01-02
Changed const to use the same syntax class as var and let.

StevenTang166:
### Description of the Change

Changed the class assigned to 'const' to use the same syntax class as 'var', and 'let'. 'const' formerly had the modifier class. 

### Alternate Designs

No other alternative designs were discussed in the issue. Was chosen because 'const' should have the same scopes, because it fills the same lexical role as 'let' and 'var' do - named data assignment.

### Benefits

'const', 'let', and 'var' will have the same syntax highlighting.

### Possible Drawbacks

Since the change is quite minor, their should not be any drawbacks.

### Applicable Issues

Fixes: [https://github.com/atom/language-javascript/issues/471](https://github.com/atom/language-javascript/issues/471)

###Image Before Change

Change

###Image After Change:

Change

### DIFF
--- a/grammars/javascript.cson
+++ b/grammars/javascript.cson
@@ -879,7 +879,7 @@
     'begin': '(?<!\\.)\\b(const)(?!\\s*:)\\b'
     'beginCaptures':
       '1':
-        'name': 'storage.type.js'
+        'name': 'storage.type.const.js'
     'end': '(\\bof\\b|\\bin\\b)|(;)|(=)|(?<![,{])\\n'
     'endCaptures':
       '1':
--- a/grammars/javascript.cson
+++ b/grammars/javascript.cson
@@ -879,7 +879,7 @@
     'begin': '(?<!\\.)\\b(const)(?!\\s*:)\\b'
     'beginCaptures':
       '1':
-        'name': 'storage.modifier.js'
+        'name': 'storage.type.js'
     'end': '(\\bof\\b|\\bin\\b)|(;)|(=)|(?<![,{])\\n'
     'endCaptures':
       '1':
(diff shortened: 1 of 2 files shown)


grammars/javascript.cson
```diff
@@ -879,7 +879,7 @@
     'begin': '(?<!\\.)\\b(const)(?!\\s*:)\\b'
     'beginCaptures':
       '1':
-        'name': 'storage.modifier.js'
+        'name': 'storage.type.js'
```
MaximSokolov: It should be `storage.type.const.js` for consistency with `storage.type.var.js`

StevenTang166: Okay.

MaximSokolov:
Tests also should be updated

StevenTang166:
Am I supposed to do that, and if so how?

MaximSokolov:
Specs are in the [spec/javascript-spec.coffee](https://github.com/atom/language-javascript/blob/master/spec/javascript-spec.coffee) file. To run them use `Run package specs` command in Atom (I think the default keymap binding is `Ctrl-Alt-P`) Let me know if you need help

StevenTang166:
I have just run the package specs command, the program has four failures, and this error:
`Expected {value: 'const',scopes['source.js', 'storage.type.const.js']} to equal {value: 'const',scopes['source.js', 'storage.modifier.js']}`
What am I supposed to do now?

MaximSokolov:
Change all those `storage.modifier.js` to `storage.type.const.js`

StevenTang166:
I forgot their was more than one test. All of them have changed now. And Travis now reports no failures.

50Wliu:
Thanks!

PR #482, opened 2017-01-14
Add support for namespaced default values

Alhadis:
Both of these lines are valid JSDoc constructs:

~~~js
/**
 * @param {Object} [name=Default.value]
 * @param {Object} [name=defaultValue]
 */
~~~

A `@param` tag's default-value [accepts anything](http://usejsdoc.org/tags-param.html#multiple-types-and-repeatable-parameters) that a [`@type`](http://usejsdoc.org/tags-type.html) parameter does, so whatever's valid as a `@type` is also valid as a default value:

~~~js
/**
 * @type {Default.value}
 * @type {defaultValue}
*/
~~~

### DIFF
--- a/grammars/javascript.cson
+++ b/grammars/javascript.cson
@@ -2062,10 +2062,10 @@
                   =                                     # [foo=bar] Default parameter value
                   \\s*
                   (?:
-                    [\\w$\\s]* |                        # [foo=bar] Unquoted
-                    "[^"]*"    |                        # [foo="bar"] Double-quoted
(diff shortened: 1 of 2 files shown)


Alhadis:
@50Wliu Any thoughts on adding JavaScript highlighting to inline `@examples` inside JSDoc comments?

It'd be limited to captured regions only, so there's no risk of a `begin/end` pattern slipping past a terminating `*/` boundary and ruining the rest of the document's highlighting.

50Wliu:
Sure.

PR #483, opened 2017-01-16
Add SJS (server-side JavaScript) to filetypes

mozfreddyb:
### Requirements

I'm working a lot with SJS files. These are JS files that include some additional proprietary APIs that run on the server-side (predating NodeJS). See https://developer.mozilla.org/en-US/docs/Mozilla/httpd.js/HTTP_server_for_unit_tests#SJS:_server-side_scripts for more

### Description of the Change

Added `sjs` to the list of supported filetypes

### Alternate Designs

N/A

### Benefits

Package supports more types of jS

### Possible Drawbacks

sjs is not very widely used. It might overlap with other files that people use in atom?

### Applicable Issues

N/A

### DIFF
--- a/grammars/javascript.cson
+++ b/grammars/javascript.cson
@@ -13,6 +13,7 @@
   'jspre'
   'pac'
   'pjs'
+  'sjs'
   'xsjs'
   'xsjslib'
 ]


Alhadis:
Distribution looks okay; a [GitHub search](https://github.com/search?q=extension%3Asjs+NOT+nothack&type=Code) tosses up 53,137 results.

(Sorry, I'll get back to fixing File-Icons now, haha)

mozfreddyb:
Ah, good find. Syntactically most (but not all) of them look like JavaScript.

(No need to feel sorry at all! I'm glad someone is looking at my bug reports.)

Alhadis:
You can [use this](https://github.com/Alhadis/Harvester/blob/js/harvester.js) if you'd like to harvest a list of search results (it'll result in a list of URLs that you can run through `wget`).

I'd make it more direct, but GitHub don't make harvesting files of a particular extension very easy.

Don't worry, you won't get in trouble for using it. We use it at [Linguist](https://github.com/github/linguist) for gauging in-the-wild usage of new language definitions. =)

50Wliu:
It looks like the majority are JavaScript, with a few binary files.  So I think this is fine.

PR #488, opened 2017-02-06
Added space in function definitions

PierBover:
To be more consistent with the other snippets in this package I've added a space in some function definitions that lacked it.

For example `if (...) {` and `while (...) {` have a space between the keyword and the parenthesis but some function definitions did not and resulted in `function foo() {`.

This behaviour is inconsistent. For example some function definitions did have the space such as:

```
  'Object Method':
    'prefix': 'kf'
    'body': '${1:methodName}: function (${2:attribute}) {\n\t$3\n}${4:,}'
```

Or also:

```
  'Prototype':
    'prefix': 'proto'
    'body': '${1:ClassName}.prototype.${2:methodName} = function ($3) {\n\t$0\n};'
```

But not in:

```
  'Function':
    'prefix': 'fun'
    'body': 'function ${1:functionName}($2) {\n\t$0\n}'
```

With the new changes all function definitions get the space so the result is `function foo () {` and `function () {`.

### DIFF
--- a/snippets/language-javascript.cson
+++ b/snippets/language-javascript.cson
@@ -37,7 +37,7 @@
     'body': 'for (${1:variable} of ${2:iterable}) {\n\t$3\n}'
   'Function':
     'prefix': 'fun'
-    'body': 'function ${1:functionName}($2) {\n\t$0\n}'
+    'body': 'function ${1:functionName} ($2) {\n\t$0\n}'
   'Anonymous Function':
     'prefix': 'f'
     'body': 'function ($1) {\n\t$2\n}'
@@ -46,7 +46,7 @@
     'body': '($1) => {\n\t$2\n}'
   'Generator':
     'prefix': 'gen',
-    'body': 'function* ${1:functionName}($2) {\n\t$0\n}'
+    'body': 'function* ${1:functionName} ($2) {\n\t$0\n}'
   'Anonymous Generator':
     'prefix': 'g'
     'body': 'function* ($1) {\n\t$2\n}'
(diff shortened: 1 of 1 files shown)


PierBover:
Any comments @50Wliu or @lee-dohm ?

lee-dohm:
Thanks for the suggestion! This is changing a default, so people that are using the current snippets and like them the way they are (or gotten used to them) will have their workflow broken. They'll have to create new snippets to maintain their current workflow. We tend to be very careful around changing the defaults on people, so there is already a high bar here.

Regarding consistency, `if` and `while` aren't function definitions, so I don't consider them to be evidence of a consistency problem. "Object Method" and "Prototype" are both assigning anonymous functions to a name, rather than creating named functions, so structurally they're different. The only place that I have seen the "function name separate from the parameter list" enforced is in [standard.js](http://standardjs.com/) and while standard.js is used by Atom and Electron and somewhat common besides, it isn't universal.

I don't feel there's enough motivation here to justify changing a default. I'll pass it by the team and see if anyone else has any input first though.

PierBover:
Thanks @lee-dohm  I understand. Yeah maybe I'm being too OCD :)

Let's see if anyone on the team agrees.

lee-dohm:
Thanks for understanding! We're all pedantic in our own ways :laughing:

PierBover:
Hey @lee-dohm I'm guessing I should just close this PR, right?

lee-dohm:
@PierBover Yes, thank you. My apologies for not getting back to you on it.

lee-dohm:
Thanks very much for the effort you put in to this @PierBover :heart:

PR #489, opened 2017-02-07
Fixed 'iife' snippet, moving final paren before function invocation

gopsmith:
This change corrects the 'iife' snippet body. The final parenthesis was in the wrong place: instead of closing the entire expression in the body, it should have been placed before the () function invocation so that the anonymous function gets invoked properly.


50Wliu:
Both the existing snippet and the proposed snippet do the same thing.  See http://benalman.com/news/2010/11/immediately-invoked-function-expression/.

PR #496, opened 2017-03-03
Decouple JSDoc from JavaScript grammar / Highlight examples

### LINKED ISSUE #495: Catastrophic backtracking in JSDoc param types
### Prerequisites

### Description

There is currently catastrophic backtracking occurring when we we fail to match the param type when multiple are present in JSDoc comments.
```js
/**
 * @param {String|String|String|String|String|String|String|String*} test 
 */
```
should be enough to reproduce.  Another example is
```js
/**
 * @param {PIXI.DisplayObject|PIXI.Container|PIXI.BaseTexture|PIXI.Texture|PIXI.Graphic*} item - Object to
 *        add to the queue
```

Thanks to @andrewstart for reporting.

### Steps to Reproduce

1. Copy/paste one of the two snippets into a new editor and set the syntax to JavaScript

**Expected behavior:** Instant highlighting

**Actual behavior:** Keep Waiting dialog

**Reproduces how often:** 100%

### Versions

language-javascript@0.126.1, but it's been happening since at least 0.122.0.

/cc @bgriffith @Alhadis

Alhadis:
Fixes #495.

### Description of the Change

This PR separates the patterns for highlighting [JSDoc](http://usejsdoc.org/) tags from the main JavaScript grammar. The lookahead patterns currently responsible for this are inflicting severe [performance issues](https://github.com/atom/language-javascript/issues/495) in certain cases, which is to say nothing of the maintainability issues they've always imposed.

Additionally, I've also included inline JS highlighting for `@example` tags:

Figure 1

### Alternate Designs / Thoughts

- Initially, I wrote each block tag to span multiple lines, but realised highlighting something like `@access \n private` really didn't justify the added complexity.

### Benefits

Other than being much more readable and maintainable, I imagine this will boost the tokeniser's performance drastically. It might be interesting to benchmark.

### Possible Drawbacks

In light of the countless intricacies of Closure Compiler, I've opted to match `{types}` and `@param` tags with an inclusive approach, rather than exclusive. Users will no longer be able to rely on what this grammar matches to indicate what syntax is valid. To be honest, I feel it's a fair trade-off... grammars shouldn't be responsible for validating input (at least beyond the scope of unclosed strings, etc).


50Wliu:
Oh look, another huge PR to review from @Alhadis :grin:.  I may need to punt this for a while to get a language-c rewrite merged first, as that grammar is in pretty bad shape right now.

grammars/jsdoc.cson
```diff
@@ -0,0 +1,478 @@
...
+    # Tags followed by a quoted arbitrary value
+    # -  @<tag> "Quoted value"
+    'begin': '((@)(?:default(?:value)?|license|version))\\s+([\'"])'
+    'beginCaptures':
+      '1':
+        'name': 'storage.type.class.jsdoc'
+      '2':
+        'name': 'punctuation.definition.block.tag.jsdoc'
+      '3':
+        'name': 'variable.other.jsdoc'
```
50Wliu: Quotes should get their own quoted token

grammars/jsdoc.cson
```diff
@@ -0,0 +1,478 @@
...
+      '1':
+        'name': 'punctuation.definition.bracket.square.begin.jsdoc'
+      '2':
+        'name': 'punctuation.definition.bracket.square.end.jsdoc'
+    'match': '(\\[)[^\\]]+(\\])(?={@(?:link|linkcode|linkplain|tutorial))'
+    'name': 'constant.other.description.jsdoc'
+  }
+  {
+    # {@link|tutorial …}
+    'begin': '({@)(?:link|linkcode|linkplain|tutorial)\\s*'
```
50Wliu: The bracket needs to be separated from the at-sign.  Also, the comment is a bit different from the regex match here.  It looks like the comment is specifying `{@link}` or `{tutorial}`, but the match is specifying `{@link}`, `{@tutorial}`.

grammars/jsdoc.cson
```diff
@@ -0,0 +1,478 @@
...
+          '2':
+            'name': 'punctuation.separator.pipe.jsdoc'
+        'match': '\\G((?:[^{}@\\s|*]|\\*[^/])+)(\\|)?'
+      }
+    ]
+  }
+]
+
+'repository':
+
```
50Wliu: :fire: unnecessary line

grammars/jsdoc.cson
```diff
@@ -0,0 +1,478 @@
...
+        'begin': '\\['
+        'end': '\\]|(?=\\*/)'
+        'patterns': [
+          {
+            'include': '#brackets'
+          }
+        ]
+      }
+    ]
+
```
50Wliu: :fire: empty line

grammars/jsdoc.cson
```diff
@@ -0,0 +1,478 @@
...
+        'match': '\\G((?:[^{}@\\s|*]|\\*[^/])+)(\\|)?'
+      }
+    ]
+  }
+]
+
+'repository':
+
+  # Balanced brackets (square or curly)
+  'brackets':
```
50Wliu: Should we add names for these or no?

grammars/jsdoc.cson
```diff
@@ -0,0 +1,478 @@
...
+        'begin': '"'
+        'end': '"|(?=\\*/)'
+        'patterns': [
+          {
+            'include': '#quotes'
+          }
+        ]
+      }
+    ]
+
```
50Wliu: :fire: empty line

spec/jsdoc-spec.coffee
```diff
@@ -0,0 +1,1225 @@
...
+  grammar = null
+
+  beforeEach ->
+    waitsForPromise ->
+      atom.packages.activatePackage("language-javascript")
+
+    runs ->
+      grammar = atom.grammars.grammarForScopeName("source.js")
+
+
```
50Wliu: :fire: empty line

spec/jsdoc-spec.coffee
```diff
@@ -0,0 +1,1225 @@
...
+      expect(tokens[7]).toEqual value: '*/', scopes: ['source.js', 'comment.block.documentation.js', 'punctuation.section.end.comment.js']
+
+    it "tokenises tags with an embedded trailing description", ->
+      {tokens} = grammar.tokenizeLine('/** Text {@linkplain target|Description text} */')
+      expect(tokens[0]).toEqual value: '/**', scopes: ['source.js', 'comment.block.documentation.js', 'punctuation.section.begin.comment.js']
+      expect(tokens[1]).toEqual value: ' Text ', scopes: ['source.js', 'comment.block.documentation.js']
+      expect(tokens[2]).toEqual value: '{@', scopes: ['source.js', 'comment.block.documentation.js', 'entity.name.type.instance.jsdoc', 'storage.type.class.jsdoc', 'punctuation.definition.inline.tag.begin.jsdoc']
+      expect(tokens[3]).toEqual value: 'link', scopes: ['source.js', 'comment.block.documentation.js', 'entity.name.type.instance.jsdoc', 'storage.type.class.jsdoc']
+      expect(tokens[4]).toEqual value: 'plain', scopes: ['source.js', 'comment.block.documentation.js', 'entity.name.type.instance.jsdoc', 'variable.other.description.jsdoc']
```
50Wliu: This is interesting...shouldn't it be tokenized as `linkplain` and not separately?

spec/jsdoc-spec.coffee
```diff
@@ -0,0 +1,1225 @@
...
+      expect(tokens[8]).toEqual value: '*/', scopes: ['source.js', 'comment.block.documentation.js', 'punctuation.section.end.comment.js']
+
+      {tokens} = grammar.tokenizeLine('/** Text {@linkcode target Description text} */')
+      expect(tokens[0]).toEqual value: '/**', scopes: ['source.js', 'comment.block.documentation.js', 'punctuation.section.begin.comment.js']
+      expect(tokens[1]).toEqual value: ' Text ', scopes: ['source.js', 'comment.block.documentation.js']
+      expect(tokens[2]).toEqual value: '{@', scopes: ['source.js', 'comment.block.documentation.js', 'entity.name.type.instance.jsdoc', 'storage.type.class.jsdoc', 'punctuation.definition.inline.tag.begin.jsdoc']
+      expect(tokens[3]).toEqual value: 'link', scopes: ['source.js', 'comment.block.documentation.js', 'entity.name.type.instance.jsdoc', 'storage.type.class.jsdoc']
+      expect(tokens[4]).toEqual value: 'code', scopes: ['source.js', 'comment.block.documentation.js', 'entity.name.type.instance.jsdoc', 'variable.other.description.jsdoc']
```
50Wliu: Ditto

spec/jsdoc-spec.coffee
```diff
@@ -0,0 +1,1225 @@
...
+      expect(tokens[10]).toEqual value: ' parameter.property = default value ', scopes: ['source.js', 'comment.block.documentation.js', 'variable.other.jsdoc']
+      expect(tokens[11]).toEqual value: ']', scopes: ['source.js', 'comment.block.documentation.js', 'variable.other.jsdoc']
+      expect(tokens[12]).toEqual value: ' this is the description ', scopes: ['source.js', 'comment.block.documentation.js']
+
+    it "tokenises @param tags with wildcard types", ->
+      {tokens} = grammar.tokenizeLine('/** @param {*} variable this is the description */')
+      expect(tokens[5]).toEqual value: '{', scopes: ['source.js', 'comment.block.documentation.js', 'entity.name.type.instance.jsdoc', 'punctuation.definition.bracket.curly.begin.jsdoc']
+      expect(tokens[6]).toEqual value: '*', scopes: ['source.js', 'comment.block.documentation.js', 'entity.name.type.instance.jsdoc']
+      expect(tokens[7]).toEqual value: '}', scopes: ['source.js', 'comment.block.documentation.js', 'entity.name.type.instance.jsdoc', 'punctuation.definition.bracket.curly.end.jsdoc']
+      expect(tokens[9]).toEqual value: 'variable', scopes: ['source.js', 'comment.block.documentation.js', 'variab
```
50Wliu: :fire: empty line

50Wliu:
Would it be worthwhile to tokenize default values as actual JS?

grammars/jsdoc.cson
Alhadis: This is only used to keep stuff balanced inside default values. I chose not to add any scopes to their contents so it was clearer to the reader where the JSDoc token ended.

grammars/jsdoc.cson
```diff
@@ -0,0 +1,478 @@
...
+        # URL
+        (
+          (?=https?://)
+          (?:[^\\s*]|\\*[/])+
+        )
+        |
+        # JSDoc namepath
+        (
+          (?!https?://)
+          (?:[^@\\s*/]|\\*[^/])++
```
Alhadis: Because I'm a possessive prick.

`(removed)*+`

grammars/jsdoc.cson
```diff
@@ -0,0 +1,478 @@
...
+      '4':
+        'name': 'entity.name.type.instance.jsdoc'
+    'match': '''(?x)
+      ((@)see)
+      \\s+
+      (?:
+        # URL
+        (
+          (?=https?://)
+          (?:[^\\s*]|\\*[/])+
```
Alhadis: Good catch, fixed.

Alhadis:
Possibly, but that would likely leave us at the mercy of runaway `begin`/`end` pairs if the terminating `]` was matched by something else.

Alhadis:
... never mind, I said nothing.

I shouldn't be allowed to communicate with humans until I've finished waking up.

grammars/jsdoc.cson
```diff
@@ -0,0 +1,502 @@
...
+        'name': 'constant.other.email.link.underline.jsdoc'
+      '6':
+        'name': 'punctuation.definition.bracket.angle.end.jsdoc'
+  }
+  {
+    # @borrows <that namepath> as <this namepath>
+    'match': '''(?x)
+      ((@)borrows) \\s+
+      ((?:[^@\\s*/]|\\*[^/])+)    # <that namepath>
+      \\s++ (as) \\s+             # as
```
50Wliu: You forgot to remove the first possessive match on this line.

grammars/jsdoc.cson
```diff
@@ -0,0 +1,502 @@
...
+      '2':
+        'name': 'punctuation.definition.block.tag.jsdoc'
+    'end': '(?=\\s|\\*/|[^{}\\[\\]A-Za-z_$])'
+    'patterns': [
+      {
+        'include': '#type'
+      }
+      {
+        'name': 'entity.name.type.instance.jsdoc'
+        'match': '(?:[^@\\s*/]|\\*[^/])++'
```
50Wliu: Possessive

grammars/jsdoc.cson
```diff
@@ -0,0 +1,502 @@
...
+    '''
+  }
+  {
+    # @author name [<email>]
+    'match': '''(?x)
+      ((@)author)
+      \\s+
+      (
+        [^@\\s<>*/]
+        (?:[^@<>*/]|\\*[^/])*+
```
50Wliu: Possessive

grammars/jsdoc.cson
```diff
@@ -0,0 +1,502 @@
...
+          '0':
+            'name': 'entity.name.tag.inline.jsdoc'
+          '1':
+            'name': 'punctuation.definition.bracket.angle.begin.jsdoc'
+          '2':
+            'name': 'punctuation.definition.bracket.angle.end.jsdoc'
+      }
+      {
+        # Highlighted JavaScript example
+        'match': '[^\\s@*](?:[^*]|\\*[^/])*+'
```
50Wliu: Possessive

grammars/jsdoc.cson
```diff
@@ -0,0 +1,502 @@
...
+      ((@)author)
+      \\s+
+      (
+        [^@\\s<>*/]
+        (?:[^@<>*/]|\\*[^/])*+
+      )
+      (?:
+        \\s*
+        (<)
+        ([^>\\s]++)
```
50Wliu: Possessive

Alhadis:
Fixed.

Alhadis:
@50Wliu Urgh, I forgot to accommodate arrays when highlighting default values as JavaScript. Something like this will only be partially highlighted:

~~~js
/** @param {Object} [thing={a: [], b: [0, 2], c: "String"}] [Not Highlighted] [] [] [] */
~~~

Currently working on a fix, although this is surprisingly tricky to get right...

50Wliu:
If it's too difficult/tricky to achieve just revert the embedded highlighting all together.

Alhadis:
No no, so far, so good... it's just that if users get too carried away with really convoluted complex values, they might get flaky-looking highlighting.

Then again, if they're stuffing that much into a default value, what else do you expect? 😀 Heck, this might even discourage an anti-pattern, haha.

Alhadis:
Okay, fixed. Pleeeease don't chide me over the odd-looking way I've implemented this... :D Some repetition *was* necessary...

PR #497, opened 2017-03-10
Improve highlighting of embedded @example source

Alhadis:
Follow-up of [an issue I noticed after the responsible PR was merged](https://github.com/atom/language-javascript/pull/496#issuecomment-285577005).

**Cons:** It's left the patterns looking hairier than I am.

**Pros:** The output looks good enough to eat:

Figure 1 & 2

- **Top:** My hand-rolled, deliberately-plain and undistracting syntax theme
- **Bottom:** The beautifully-elegant, almost palpably paper-like [DuoTone Light](https://atom.io/themes/duotone-light-syntax)

-------------------

**EDIT 14/03:** Okay, a few more additions... I feel bad piling them onto an unrelated PR, but I figure it'll probably get things reviewed quicker (since this already needs revisiting).

1. HTML highlighting added to embedded `<caption>` tags:

Figure 3

2. `<caption>` tags will now match on a newline following the opening `@example` tag

3 `@api` and `@internal` added to recognised JSDoc tags; [see comment below](https://github.com/atom/language-javascript/pull/497#issuecomment-285845222)


Alhadis:
@50Wliu I also noticed JSDoc drops any text that's touching the closing bracket of a default @param value, so I've highlighted it as an error:

Figure 2

Specs included, of course. =)

Alhadis:
Okay, aaand I noticed incomplete highlighting of `@see {@link http://url.com/}` (only the opening `{` was being highlighted). I remember I meant to get back to that... sorry! :( I've added a fix for it to this PR.

Alhadis:
@50Wliu Is it okay to add `@api` and `@internal` as highlighted tags? They're not part of JSDoc, but they are used in-the-wild for similar API-generation systems.

E.g.:
* [`@internal`](https://github.com/ng-bootstrap/ng-bootstrap/blob/master/src/accordion/accordion.ts#L198)
* [`@internal`](https://github.com/palantir/blueprint/wiki/Coding-guidelines) (Grep`"@internal" in the page)
* [`@api`](https://github.com/mochajs/mocha/blob/master/lib/runnable.js#L44) (All throughout MochaJS's codebase)

grammars/jsdoc.cson
```diff
@@ -252,28 +275,29 @@
         'name': 'variable.other.jsdoc'
       }
       {
-        'name': 'variable.other.jsdoc'
         'begin': '\\['
-        'end': '\\]|(?=\\*/)'
+        'beginCaptures':
+          '0':
+            'name': 'variable.other.jsdoc'
+        'contentName': 'variable.other.jsdoc'
```
50Wliu: Use `name` here instead of `contentName` if the begin and end captures should get the same scopes.

grammars/jsdoc.cson
```diff
@@ -411,6 +435,120 @@
         ]
       }
     ]
+  # Default @param value, highlighted as JavaScript
+  'default-value':
+    'patterns': [
+      # Double-quoted string
+      {
+        'begin': '(")'
```
50Wliu: Why are we special-casing strings?

grammars/jsdoc.cson
```diff
@@ -411,6 +435,120 @@
...
+        'name': 'source.embedded.js'
+        'patterns': [
+          {
+            'match': '\\\\(?:[^*]|\\*(?!/))'
+            'name': 'constant.character.escape.js'
+          }
+        ]
+      }
+      {
+        # Nested arrays matched to keep brackets balanced
```
50Wliu: What exactly is this for?

grammars/jsdoc.cson
Alhadis: It's to keep `invalid.illegal.syntax.jsdoc` from being nested inside `variable.other.jsdoc`. For themes that prioritise the former over the latter, we don't want invalid text being highlighted as valid.

grammars/jsdoc.cson
Alhadis: To stay consistent with JavaScript. Furthermore, some themes might colour double-quoted strings differently to single-quoted strings ([such as `seti-syntax`](https://github.com/jesseweed/seti-syntax/blob/29fc4f4/styles/syntax.less#L347-L350)).

grammars/jsdoc.cson
Alhadis: Stuff like `@param {Object} [value = {key: [ [0], [1] ] }] Description`

grammars/jsdoc.cson
Alhadis: In a theme which doesn't highlight `invalid.illegal.syntax.jsdoc`:

Alhadis:
@50Wliu Any chance we could get this looked at soon?

Linguist is probably due for another release soon, so it'd be great to get this working across GitHub...

50Wliu:
I was actually reminding myself to look at this again last night, also because a new Atom release is approaching.

I have mostly refrained from doing so because it's a surprisingly large PR, which I haven't had the appetite to review.  I'll try to get to it soon, but like usual, no guarantees.

Alhadis:
If it helps, I kept my [last commits atomic and documented](https://github.com/atom/language-javascript/pull/497/commits) to ameliorate the hassle. =)

grammars/jsdoc.cson
50Wliu: I don't like this.  I feel like it introduces complexities in the grammar, and I've noticed that while it's unnoticeable while _writing_ code, internally the scopes are still separated (such as in specs).

As you know I really value correct highlighting, but I think in this case it's better to use `name`.

grammars/jsdoc.cson
50Wliu: Same thing here.  The double `string.quoted.double.js` capture/contentName feel _extremely_ hacky to me, and is actually why it's taken me so long to get back to reviewing this PR.

Is there any other way to do this without the double capture?

Alhadis: Would you prefer I simply embed `source.js` in the captured match, then? Unless you want *no* string highlighting whatsoever...

grammars/jsdoc.cson
Alhadis: As you wish.

grammars/jsdoc.cson
Alhadis: Wait, scratch that suggestion. That's only gonna make things even more complex, because care will need to be taken to stop `source.js` from consuming the closing quote-character.

Is there any logical, **practical** reasoning for this, other than the old *"it seems too hacky to me"*...? Heck, for all you know, you could be ditching this grammar in six months in favour of Tree-Sitter or Bush-Croucher or something... =)

Alhadis: FYI, this isn't that different from the hacks I used in [`atom/language-xml#54`](https://github.com/atom/language-xml/pull/54) to circumvent runaway highlighting with `</script>` tags and stuff. Heck, that PR's hacks are a hell of a lot worse, and the results aren't even gonna be appreciated on the level that consistent JSDoc highlighting will.

50Wliu: That also happens to be why I haven't finished reviewing the XML PR.  This probably won't make it into either the linguist or Atom releases as I'll need time to reflect and think about your questions.

And also, would it be feasible to include all of source.js in here after https://github.com/atom/first-mate/pull/90 is merged?

Alhadis: @50Wliu If it's really this much of an issue, I'll drop it for you. I'm not going to let this be delayed on something this trivial.

Hold on.

Alhadis: Alright, it's done. I've merged both string-matching blocks, and used `string.quoted.js` for the applied scope.

If there's anything else that's blocking this, please speak up.

Alhadis:
@50Wliu Any chance of this making the next Atom release now that I've simplified the pattern/scope-matching?

50Wliu:
I will take a look on Monday.

Alhadis:
How's Monday going?

It's Tuesday here, hahah. Was already Monday here when you posted that, too. 😀

grammars/jsdoc.cson
50Wliu: So will this be needed after language-javascript switches to using `begin`/`end` captures for square brackets?

grammars/jsdoc.cson
```diff
@@ -411,6 +430,92 @@
...
+      # Chunk of JS source (excluding nested arrays; see above). Note we "clamp" the
+      # imported highlighting within a fixed range to stop runaway problems with any
+      # begin/end patterns the JavaScript grammar matches.
+      {
+        'match': '(?:[^"\'\\[\\]*]|\\*[^/])+'
+        'name': 'source.embedded.js'
+        'captures':
+          '0':
+            'patterns': [
+              # Unmatched } after a string literal: `"string"}`
```
50Wliu: Curious what this is for.

Alhadis: Couldn't figure out what was causing it, and it was giving me the shits, so I added a different rule.

spec/jsdoc-spec.coffee
```diff
@@ -1162,55 +1370,126 @@ describe "JSDoc grammar", ->
...
-      expect(lines[3][2]).toEqual value: 'return', scopes: ['source.js', 'comment.block.documentation.js', 'storage.type.class.jsdoc']
-      expect(lines[3][4]).toEqual value: '{', scopes: ['source.js', 'comment.block.documentation.js', 'entity.name.type.instance.jsdoc', 'punctuation.definition.bracket.curly.begin.jsdoc']
-      expect(lines[3][5]).toEqual value: 'String', scopes: ['source.js', 'comment.block.documentation.js', 'entity.name.type.instance.jsdoc']
-      expect(lines[3][6]).toEqual value: '}', scopes: ['source.js', 'comment.block.documentation.js', 'entity.name.type.instance.jsdoc', 'punctuation.definition.bracket.curly.end.jsdoc']
-      expect(lines[3][7]).toEqual value: '', scopes: ['source.js', 'comment.block.documentation.js']
-      expect(lines[4][1]).toEqual value: '*/', scopes: ['source.js', 'comment.block.documentation.js', 'punctuation.section.end.comment.js']
-
+    describe "HTML captions", ->
+      ###
+        NOTE: Loading the HTML grammar triggers an inexplicable case of infinite recursion during the following two specs.
```
50Wliu: Hmm.  That is quite odd.

50Wliu:
9:07pm over here :grinning:.  I'm going to take a look at the HTML recursion.

grammars/jsdoc.cson
Alhadis: Yes, because it's used to balance all that weird, noisy syntax Google use for Closure Compiler tags, which combine `< [ {` brackets with nesting and all sorts of weird-looking rubbish.

50Wliu:
Also: I think I'm going to change the update schedule for grammars in Atom.  Instead of updating all the grammars right before a new release, I'm going to update them right _after_, so that they get the full benefit of Atom's release pattern.  That also means that there will be very minimal language updates for 1.17.

Alhadis:
Eh... okay...

Wasn't the first buggy JSDoc PR I submitted merged in a release used by the next Atom version?

50Wliu:
Nope: #496 hasn't been included in a language-javascript release yet.

Also, I can reproduce the infinite recursion in a production environment.  The repro case is as simple as having the caption element, as in the first spec, and then trying to disable and then re-enable language-html.

Alhadis:
> Nope: #496 hasn't been included in a language-javascript release yet.

Well, thankfully Linguist doesn't need a tagged release to pick up the latest changes to a grammar. :D

As for Atom, I'm not fussed about getting this shipped soon. I have enough to worry about with keeping File-Icons stable until the 5 PRs are dealt with.

Alhadis:
@50Wliu There's **very** likely going to be a new Linguist release soon. What else is there that I have to do?

50Wliu:
The infinite recursion issue with language-html needs to be fixed.

Alhadis:
Would you prefer I remove the embedded HTML highlighting for the time being, then?

50Wliu:
I'm not opposed to that.  It can always be added later.

Alhadis:
Precisely my line of thinking as well. Hold on.

Alhadis:
Righto, done. =)

Damn strange I can't reproduce the recursion error outside of the spec-runner, though.

50Wliu:
Did you see my [reproduction steps](https://github.com/atom/language-javascript/pull/497#issuecomment-293123086) above?

Alhadis:
Ah, I missed the part about disabling and reenabling. Sorry, never mind.

50Wliu:
I will try to give a final review within the next few days.

grammars/jsdoc.cson
```diff
@@ -411,6 +413,92 @@
...
+      {
+        'begin': '("|\')'
+        'beginCaptures':
+          '0':
+            'name': 'punctuation.definition.string.begin.js'
+        'end': '\\1|(?=\\*/)'
+        'endCaptures':
+          '0':
+            'name': 'punctuation.definition.string.end.js'
+        'name': 'string.quoted.js'
```
50Wliu: I noticed that in bd688cc0c02d4e8ae094f47b6ce4b0ad692f6cc5 you simplified the string rules, but shouldn't name still differentiate between single and double quoted strings?

grammars/jsdoc.cson
50Wliu: Can you give me a code example of what caused you to add this rule?

grammars/jsdoc.cson
Alhadis: How am I expected to do that without duplicating the rule?

grammars/jsdoc.cson
Alhadis: No, sorry, I've completely forgotten what I was even working on when I encountered this.

grammars/jsdoc.cson
50Wliu: I'm fine with the rule being duplicated.  Just like how it already is in the main javascript grammar (https://github.com/atom/language-javascript/blob/f470e132e2988d6d088a4746b7a494ed78e3e06c/grammars/javascript.cson#L1338-L1375).

Alhadis: Done.

grammars/jsdoc.cson
Alhadis: Are you waiting for me to remove this rule, or...?

50Wliu: I wasn't notified of your last response.  I'll see if I can figure out why the curly brackets aren't being matched properly.

Alhadis: It's possible I added it before another pattern was revised (which fixed the same issue). If you can't reproduce it, don't sweat it. :) I'm happy to shoot the damn thing if it moves things along.

50Wliu: So it quickly became obvious after I tried removing it and running specs.  Given:
```js
/** @param {Object} [variable={a: "b"}] - An object */
```
The `default-value` pattern "clamps" (in your own words) the JS highlighting to stop at the opening quotation mark.  Therefore we start with `meta.brace.curly.js` but then when we pick up JS highlighting again after the closing quotation we can't end it.

Alhadis: Ugh, of course. Now I remember: I was making sure the JS highlighting wouldn't consume any more than it was supposed to. Just erring on the side of caution. =)

I guess it's acceptable to leave, then? Or would you rather I include an inline explanation?

50Wliu: I'll tinker around and see if I can come up with a good compromise.

Alhadis: [Heads up](https://github.com/github/linguist/pull/3587#issuecomment-298137430), there's not much time. How about just putting a link to this comment thread in the grammar and wrapping this up?

Alhadis: @50Wliu May I ask why this is a problem?

50Wliu: I think you've realized by now that you and I sometimes disagree on what is shippable.  I apologize for the long delays between reviews and comments; that's something I need to work on when something like this occurs.

If you would really like Linguist to include the new JSDoc highlighting, I would encourage you to submit a (hopefully temporary) PR that eliminates the embedded JS highlighting.  Then, we can continue discussing.

I've been trying out different methods to avoid the scope-chopping that's occurring here.  I've seen some success but it'll take more time and I don't want to delay your Linguist release or have it ship with buggy highlighting.

Alhadis: > I think you've realized by now that you and I sometimes disagree on what is shippable.

What you're neglecting to do is explain to me *why*. So far, it seems like a case of "it doesn't *feel* right to me". That's what's disconcerting to me: that's why 😨

Alhadis: ... I hate phone keyboards.

* that's why I'm becoming this frustrated.

Alhadis: Once I'm back in front of a computer, I'm replacing this grammar with a fork until you've come to a resolution.

Alhadis: Right, never mind. I [explained the issue](https://github.com/github/linguist/pull/3587#issuecomment-298878954) to Linguist's maintainer and he believes the best the thing to do at this point is to avoid updating the JavaScript grammar with the next release, so the grammar's current state won't affect GitHub.

Fix this whenever. No hurry now.

50Wliu: Yeah.  So I didn't explain why because it was getting late last night but I still wanted to respond to let you know I got the notification.

Here's why I think this PR is unmergeable as-is, or even with an explanatory comment.
* These changes result in certain parts of language-javascript being spliced up such that the pattern begins matching but is interrupted before the end match.
* Therefore you had to add some out-of-context rules in an attempt to balance out the mismatched rules.  This is very prone to breakage as scopes or matches change.
* Inconsistent scopes.  Let's look at https://github.com/atom/language-javascript/pull/497/files#diff-c4166255a13782b6557ec1da0f4a6ef3R467 for an example.  Why is `"b"` not tokenized as embedded JavaScript, while it should be?  The whole default value should be embedded JS, yet there are portions when that scope disappears.  This of course is partly due to my insistence that double-matching should be avoided where possible, but that also brings me to my next point:
* Duplication of rules found in javascript.cson.  Again, this is very hard to keep in sync.
* Considerably, as you described it, "hairier" patterns for comparably small highlighting gains.

Overall, after reviewing this PR I am unconvinced that embedded Javascript highlighting for default values is worth it if it cannot be done in a clean manner that is easy to understand and requires minimal maintenance upkeep.

Alhadis:
I won't waste your time anymore. Sorry.

Alhadis:
Going forward, I think the best thing for me to do is refrain from submitting PRs to grammar packages, as it's obvious our opinions of grammar structure are too different for these things to be tolerable for either of us.

PR #512, opened 2017-05-24
Re-implement embedded JSDoc JavaScript highlighting

50Wliu:
### Requirements

### Description of the Change

Implements #497 using a single match instead of multiple begin/ends.  Taken from the previous JSDoc implementation.  Regexes are still a bit ugly, but less hacky than #497.

### Alternate Designs

:fire: embedded JS highlighting in JSDoc as a whole.

### Benefits

Arrays in JSDoc optional values should receive proper JavaScript highlighting.

### Possible Drawbacks

Messy regexes = large potential for regressions.  @Alhadis wrote some wonderful tests though.

### Applicable Issues

#497

/cc @Alhadis - would you mind reviewing this?  I'm still working on HTML caption support but thought that this was a good place to request a review.


Alhadis:
Sorry for not hearing your ping, mate. Crippling depression, etc, etc... I've probably missed other things at other repositories, too. :(

50Wliu:
Don't think so.  This was the only one I specifically pinged you on.

PR #515, opened 2017-06-03
Add support for multi-line types in jsdoc.

ribrdb:
### Requirements

### Description of the Change

Support for multi line types in jsdoc.
See https://github.com/Microsoft/TypeScript-TmLanguage/issues/467

And here's an example file with multi-line types:
https://github.com/google/closure-library/blob/master/closure/goog/dom/animationframe/animationframe.js#L61

### Alternate Designs

I don't know of any.

### Benefits

Multi line types are correctly highlighted.

### Possible Drawbacks

Unterminated types are not recognized as invalid. I couldn't figure out any way to maintain that. However you can clearly see that the type runs on longer than you're expecting if you forget the closing }, so I don't think that's a major issue.

### DIFF
--- a/grammars/jsdoc.cson
+++ b/grammars/jsdoc.cson
@@ -410,6 +410,10 @@
           }
         ]
       }
+      {
+        'match': '^\\s*\\*(?!/)'
+        'name': 'comment.block.documentation.jsdoc'
+      }
     ]
   'inline-tags':
     'patterns': [
@@ -483,11 +487,6 @@
   # {type}
   'type':
     'patterns': [
-      {
-        # {unclosed
-        'match': '\\G{(?:[^}*]|\\*[^/}])+$'
-        'name': 'invalid.illegal.type.jsdoc'
-      }
       {
         'begin': '\\G({)'
         'beginCaptures':
--- a/spec/jsdoc-spec.coffee
+++ b/spec/jsdoc-spec.coffee
@@ -936,6 +936,36 @@ describe "JSDoc grammar", ->
       expect(tokens[7]).toEqual value: '}', scopes: ['source.js', 'comment.block.documentation.js', 'entity.name.type.instance.jsdoc', 'punctuation.definition.bracket.curly.end.jsdoc']
       expect(tokens[9]).toEqual value: 'variable', scopes: ['source.js', 'comment.block.documentation.js', 'variable.other.jsdoc']
 
+      {tokens} = grammar.tokenizeLine('/** @param {\nnumber\n} variable this is the description */')
+      expect(tokens[5]).toEqual value: '{', scopes: ['source.js', 'comment.block.documentation.js', 'entity.name.type.instance.jsdoc', 'punctuation.definition.bracket.curly.begin.jsdoc']
+      expect(tokens[6]).toEqual value: '\nnumber\n', scopes: ['source.js', 'comment.block.documentation.js', 'entity.name.type.instance.jsdoc']
+      expect(tokens[7]).toEqual value: '}', scopes: ['source.js', 'comment.block.documentation.js', 'entity.name.type.instance.jsdoc', 'punctuation.definition.bracket.curly.end.jsdoc']
+      expect(tokens[9]).toEqual value: 'variable', scopes: ['source.js', 'comment.block.documentation.js', 'variable.other.jsdoc']
+
+      {tokens} = grammar.tokenizeLine('/** @param {{number}} variable this is the description */')
+      expect(tokens[5]).toEqual value: '{', scopes: ['source.js', 'comment.block.documentation.js', 'entity.name.type.instance.jsdoc', 'punctuation.definition.bracket.curly.begin.jsdoc']
+      expect(tokens[6]).toEqual value: '{', scopes: ['source.js', 'comment.block.documentation.js', 'entity.name.type.instance.jsdoc']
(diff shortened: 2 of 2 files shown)


50Wliu:
I think it's worth tokenizing `{{` as a _single_ `punctuation.definition.bracket.curly.begin.jsdoc` (same for ending brackets), and maybe even applying specific rules when that is encountered.  I believe that should allow us to maintain the invalid highlighting.  Thoughts?

spec/jsdoc-spec.coffee
```diff
@@ -936,6 +936,36 @@ describe "JSDoc grammar", ->
       expect(tokens[7]).toEqual value: '}', scopes: ['source.js', 'comment.block.documentation.js', 'entity.name.type.instance.jsdoc', 'punctuation.definition.bracket.curly.end.jsdoc']
       expect(tokens[9]).toEqual value: 'variable', scopes: ['source.js', 'comment.block.documentation.js', 'variable.other.jsdoc']
 
+      {tokens} = grammar.tokenizeLine('/** @param {\nnumber\n} variable this is the description */')
```
50Wliu: Instead of using `\n`, could you please use `grammar.tokenizeLines` and multiline quotes?  See https://github.com/atom/language-javascript/blob/master/spec/jsdoc-spec.coffee#L1300 for an example.

ribrdb:
{{ seems to be the most common reason to split a type across lines, but I don't think it's the only one. For example a function type with many arguments or a union type with many alternatives can also get long.

grammars/jsdoc.cson
```diff
@@ -410,6 +410,10 @@
           }
         ]
       }
+      {
+        'match': '^\\s*\\*(?!/)'
+        'name': 'comment.block.documentation.jsdoc'
```
nguyennangmao87: Comment. #114

ribrdb: I'm not sure what you mean. Do you want me to add a comment describing this rule? I don't see what this has to do with issue 114.

50Wliu: We occasionally get spammers due to the popularity of the Atom project.  I think this comment is safe to ignore.

Alhadis:
My approach would've been to add a separate rule for multiline types, prepended to the JSDoc grammar's `#type` rule. Something like this:

~~~cson
	'type':
		'patterns': [
			{
				# {{ …Multiline type… }}
				'name': 'meta.multiline-type.jsdoc'
				'begin': '\\G\\{\\{'
				'end':   '\\}\\}|(?=\\*/)'
				'beginCaptures': 'punctuation.section.bracket.curly.begin.jsdoc'
				'endCaptures':   'punctuation.section.bracket.curly.end.jsdoc'
				'patterns': [ '…etc…' ]
			}
			{
				# {unclosed
				'match': '\\G{(?:[^}*]|\\*[^/}])+$'
				'name': 'invalid.illegal.type.jsdoc'
~~~

Dunno exactly what patterns should go in the above example's `patterns` array, because I've never heard of multiline types until now.

But this approach, while more work, would allow us to keep both the existing error-highlights, and multiline type-definitions.

grammars/jsdoc.cson
Alhadis: Wrong scope-name. It should be `punctuation.section.continuation.comment.js`, or at least just `punctuation.section.comment.js`. [See here](https://github.com/atom/language-javascript/blob/ea43f15/grammars/javascript.cson#L1901-L1905).

But this pattern isn't enclosing any content of its own, so it's incorrect to scope it as a comment.

ribrdb:
A multi line jsdoc tag is just one that has the 2nd bracket on another line. Often it will start with '{{', but not always. The only way to tell if you've really got an invalid tag is to see if you get the */ before the closing '}'. But the grammar syntax doesn't have any way to do that, since it only does line based matching.

spec/jsdoc-spec.coffee
```diff
@@ -1145,6 +1145,52 @@ describe "JSDoc grammar", ->
       expect(tokens[7]).toEqual value: '}', scopes: ['source.js', 'comment.block.documentation.js', 'entity.name.type.instance.jsdoc', 'punctuation.definition.bracket.curly.end.jsdoc']
       expect(tokens[9]).toEqual value: 'variable', scopes: ['source.js', 'comment.block.documentation.js', 'variable.other.jsdoc']
 
+      lines = grammar.tokenizeLines("""
+        /** @param {
+              number
+            } variable this is the description */
+      """)
+      expect(lines[0][5]).toEqual value: '{', scopes: ['source.js', 'comment.block.documentation.js', 'entity.name.type.instance.jsdoc', 'punctuation.definition.bracket.curly.begin.jsdoc']
+      expect(lines[1][0].value).toMatch(/number/)
```
50Wliu: Can you change this to use the `.toEqual` style?  Same on L1177.

villzor:
Is there an update on this feature?

duckbrain:
It seems like the only thing this is waiting for is a style change from code review. Can I submit another PR with those changes to get this moving?

jkantr:
Apparently the previous comment is no longer accurate, as this needs to be migrated to Tree-sitter grammar to be accepted, is that correct?

Been sort of in limbo now for a few months where I have a feature that works, but without the syntax highlighting to go along with it.... is there help needed to get this resolved?

PR #521, opened 2017-07-14
Revert "Temporarily revert comment scope changes"

50Wliu:
Reverts atom/language-javascript#520

/cc @Alhadis: What do you think about `punctuation.definition.comment.begin/end.js` instead?


Alhadis:
Sounds fine to me.

50Wliu:
#522.

PR #528, opened 2017-08-11
Add .mjs as a supported file type

aemino:
### Description of the Change
At this point, the Node.js CTC has [settled on](https://github.com/nodejs/node-eps/blob/master/002-es-modules.md#32-determining-if-source-is-an-es-module) the `.mjs` extension for ES6 module files.

In addition, the [`@std/esm`](https://www.npmjs.com/package/@std/esm) package has been released, which allows inline, spec-compliant usage of ES6 modules in Node.js. In an effort to push the migration from `.js` to `.mjs`, the module defaults to only affecting files with the `.mjs` extension.

### Benefits
This change will allow those wishing to try out ES6 modules in Node.js today with proper syntax highlighting for their `.mjs` files.


50Wliu:
@aemino can you include it in alphabetical order please?

aemino:
Done. I didn't really notice the convention because of `js` and `_js` at the top. 😊

50Wliu:
Yeah, the two exceptions :).

PR #535, opened 2017-09-21
Fix catastrophic backtracking?

### LINKED ISSUE #533: Freezes Atom when parsing this code
### Prerequisites

### Description

Atom locks up when attempting to parse certain JavaScript

### Steps to Reproduce

1. Open a file

2. Set the language to `JavaScript`

3. Paste the following text in:
```js
/**
 * @param {array} [bar = 'x'] REMOVE THE CLOSE BRACKET HERE.
 */
```

4. Follow the instructions in the text and remove the `]` on line 2.
5. Note that Atom has locked up.

**Expected behavior:**

Atom doesn't freeze.

**Actual behavior:**

"Editor is not responding message" after the timeout finishes.

**Reproduces how often:**

100% of the time.

### Versions

Tested personally on:

```
Atom    : 1.21.0-beta1
Electron: 1.6.9
Chrome  : 56.0.2924.87
Node    : 7.4.0
apm  1.18.5
npm  3.10.10
node 6.9.5 x64
python 2.7.13
git 2.14.1.windows.1
visual studio 2015
```

Windows 10 x64

Reported initially on:
Atom: 1.20.1  
OS: macOS

### Additional Information

Originally filed by @danfuzz in https://github.com/AtomLinter/linter-eslint/issues/1028.

50Wliu:
@Alhadis please review and tell me there's a better solution than this

Fixes #533

### DIFF
--- a/grammars/jsdoc.cson
+++ b/grammars/jsdoc.cson
@@ -276,11 +276,11 @@
             \\s*
             (
               # The inner regexes are to stop the match early at */ and to not stop at escaped quotes
-              (?:
+              (?>
                 "(?:(?:\\*(?!/))|(?:\\\\(?!"))|[^*\\\\])*?" |  # [foo="bar"] Double-quoted
                 '(?:(?:\\*(?!/))|(?:\\\\(?!'))|[^*\\\\])*?' |  # [foo='bar'] Single-quoted
                 \\[ (?:(?:\\*(?!/))|[^*])*? \\] |              # [foo=[1,2]] Array literal
-                (?:(?:\\*(?!/))|[^*])*?                        # Everything else
+                (?:(?:\\*(?!/))|\\s(?!\\s*\\])|\\[.*?(?:\\]|(?=\\*/))|[^*\\s\\[\\]])* # Everything else (sorry)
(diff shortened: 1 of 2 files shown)


grammars/jsdoc.cson
```diff
@@ -276,11 +276,11 @@
             \\s*
             (
               # The inner regexes are to stop the match early at */ and to not stop at escaped quotes
-              (?:
+              (?>
```
Alhadis: I've tested this locally, and it works. 👍

However, there's no hairy backtracking being performed here, so it's quite strange this is happening...

Alhadis:
Only alternative solution I could suggest involves capturing everything up until either `$` or `(?=\*/)`, and defining patterns to be used within the capture.

Probably too complex FWIW, really.

grammars/jsdoc.cson
50Wliu: You can play around with the regex debugger at https://regex101.com/r/miqTHE/1 if you want.  The `*` here and the `*` on the everything else line can catastrophically backtrack when there's no `]` or `*/` present.

PR #539, opened 2017-11-01
Mark type contents with meta.embedded scope

mjbvz:
https://github.com/Microsoft/TypeScript-TmLanguage/issues/537

### Description of the Change

VS Code uses the meta.embedded scope to mark sections of strings or comments as actually being code instead of string or comment content. This change adds the meta embedded scope to the contents of jsdoc types

### Alternate Designs

Instead of modifying the existing `"contentName"`, it may be better to add a new scope for `meta.embedded.line.js`. I wasn't able to figure out how to do this. Let me know if you have any suggestions

### Benefits

Marker for what is code content with a jsdoc 

### Possible Drawbacks

Changes existing scopes

### Applicable Issues

https://github.com/Microsoft/TypeScript-TmLanguage/issues/537

### DIFF
--- a/grammars/jsdoc.cson
+++ b/grammars/jsdoc.cson
@@ -502,18 +502,15 @@
         'name': 'invalid.illegal.type.jsdoc'
       }
       {
+        'name': 'entity.name.type.instance.jsdoc'
         'begin': '\\G({)'
         'beginCaptures':
-          '0':
-            'name': 'entity.name.type.instance.jsdoc'
           '1':
             'name': 'punctuation.definition.bracket.curly.begin.jsdoc'
-        'contentName': 'entity.name.type.instance.jsdoc.meta.embedded'
-        'end': '((}))\\s*|(?=\\*/)'
+        'contentName': 'meta.embedded.line.js.jsdoc.type'
+        'end': '(})\\s*|(?=\\*/)'
         'endCaptures':
           '1':
-            'name': 'entity.name.type.instance.jsdoc'
-          '2':
             'name': 'punctuation.definition.bracket.curly.end.jsdoc'
         'patterns': [
           {
--- a/grammars/jsdoc.cson
+++ b/grammars/jsdoc.cson
@@ -508,7 +508,7 @@
             'name': 'entity.name.type.instance.jsdoc'
           '1':
             'name': 'punctuation.definition.bracket.curly.begin.jsdoc'
-        'contentName': 'entity.name.type.instance.jsdoc'
+        'contentName': 'entity.name.type.instance.jsdoc.meta.embedded'
         'end': '((}))\\s*|(?=\\*/)'
         'endCaptures':
           '1':
--- a/spec/jsdoc-spec.coffee
+++ b/spec/jsdoc-spec.coffee
@@ -80,7 +80,7 @@ describe "JSDoc grammar", ->
       expect(tokens[2]).toEqual value: '@', scopes: ['source.js', 'comment.block.documentation.js', 'storage.type.class.jsdoc', 'punctuation.definition.block.tag.jsdoc']
       expect(tokens[3]).toEqual value: 'param', scopes: ['source.js', 'comment.block.documentation.js', 'storage.type.class.jsdoc']
       expect(tokens[5]).toEqual value: '{', scopes: ['source.js', 'comment.block.documentation.js', 'entity.name.type.instance.jsdoc', 'punctuation.definition.bracket.curly.begin.jsdoc']
-      expect(tokens[6]).toEqual value: 'object', scopes: ['source.js', 'comment.block.documentation.js', 'entity.name.type.instance.jsdoc.meta.embedded']
+      expect(tokens[6]).toEqual value: 'object', scopes: ['source.js', 'comment.block.documentation.js', 'entity.name.type.instance.jsdoc', 'meta.embedded.line.js.jsdoc.type']
       expect(tokens[7]).toEqual value: '}', scopes: ['source.js', 'comment.block.documentation.js', 'entity.name.type.instance.jsdoc', 'punctuation.definition.bracket.curly.end.jsdoc']
       expect(tokens[9]).toEqual value: 'variable', scopes: ['source.js', 'comment.block.documentation.js', 'variable.other.jsdoc']
       expect(tokens[10]).toEqual value: ' - this is a ', scopes: ['source.js', 'comment.block.documentation.js']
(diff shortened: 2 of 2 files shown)


grammars/jsdoc.cson
```diff
@@ -508,7 +508,7 @@
             'name': 'entity.name.type.instance.jsdoc'
           '1':
             'name': 'punctuation.definition.bracket.curly.begin.jsdoc'
-        'contentName': 'entity.name.type.instance.jsdoc'
+        'contentName': 'entity.name.type.instance.jsdoc.meta.embedded'
```
50Wliu:
```cson
'name': 'entity.name.type.instance.jsdoc'
'contentName': 'meta.embedded.js'
```

grammars/jsdoc.cson
```diff
@@ -508,7 +508,7 @@
             'name': 'entity.name.type.instance.jsdoc'
```
50Wliu: Remove this capture

grammars/jsdoc.cson
```diff
@@ -508,7 +508,7 @@
...
           '1':
             'name': 'punctuation.definition.bracket.curly.begin.jsdoc'
-        'contentName': 'entity.name.type.instance.jsdoc'
+        'contentName': 'entity.name.type.instance.jsdoc.meta.embedded'
         'end': '((}))\\s*|(?=\\*/)'
         'endCaptures':
           '1':
```
50Wliu: Remove this capture

grammars/jsdoc.cson
```diff
@@ -508,7 +508,7 @@
             'name': 'entity.name.type.instance.jsdoc'
           '1':
             'name': 'punctuation.definition.bracket.curly.begin.jsdoc'
-        'contentName': 'entity.name.type.instance.jsdoc'
+        'contentName': 'entity.name.type.instance.jsdoc.meta.embedded'
         'end': '((}))\\s*|(?=\\*/)'
```
50Wliu: `'end': '(})|(?=\\*/)`

50Wliu:
@Alhadis is there something I'm missing with my review comments?

mjbvz:
Ok, I made the suggested change except for changing the `end` rule to remove the `\\s*`. This causes the space after the type to also be marked as `entity.name.type.instance.jsdoc` in two of the test cases but otherwise does not seem to effect things. Removing the space matching breaks the tests completely.

50Wliu:
I would prefer the solution listed [here](https://github.com/Microsoft/TypeScript-TmLanguage/issues/537#issuecomment-341246229) if possible, though this PR would still contain some nice scope fixes :).

Alhadis:
> @Alhadis Is there something I'm missing with my review comments?

Nope, you've got it right.

@mjbvz We really, really, really shouldn't remove the existing `string` and `comment` scopes for fairly obvious highlighting reasons. Imagine how jarring it'd be for an author to see JSDocs coloured like ordinary source code.

Example:

~~~js
/** "Doc-ish" comment line */
function(){

}
~~~

mjbvz:
@Alhadis Not sure I understand the part about removing scopes. The intent of this change is to add an additional `meta.embedded` scope to the type part of doc comments. VS Code looks at theses scope when deciding if editor features such as automatic suggestions should be enabled at the current position. Inside normal comments, they should not be. Inside comment content like types that is marked `meta.embedded`, we do want auto suggestions

Alhadis:
Yes, I understand. The problem is that you're assigning these scopes with equal depth: E.g., `meta.embedded.line.js.jsdoc.type`.

[TextMate's docs](http://manual.macromates.com/en/language_grammars#naming_conventions) are quite elaborate on how the `meta`-type scopes are to be used:

> The `meta` scope is generally used to markup larger parts of the document. For example the entire line which declares a function would be `meta.function` and the subsets would be `storage.type`, `entity.name.function`, `variable.parameter` etc, and only the latter would be styled.

So the author's example can be visualised as:

~~~
meta.embedded
    +
    |
    +--- entity.name.parameter
~~~

Whereas what you're doing is essentially this:

~~~
meta.embedded.entity.name.parameter
~~~

The key difference is in scope priority. Remember, the same definition of "scope" is used for two entirely seperate affairs: conveying semantic information, and syntax highlighting. You're concerned with the former, but the proposed changes are messing with the latter.

What was that Edsger Dijkstra said about separation of concerns, again...? ;)

@50Wliu I better take care of this. Both parties can be satisfied, but it takes a "text mate" to do it. 😜

Alhadis:
I should also stress this grammar has violated TextMate scope-semantics dozens of times already, and the scopes chosen for JSDoc highlighting were sadly some of the worst.  Unfortunately, what we have is what themes are using for highlighting (as well as GitHub), so we're kinda stuck with this for a while.

mjbvz:
Any updates on this? #541 does not address the same thing as this PR

50Wliu:
On my list to re-review.

50Wliu:
Sorry for the delay!  Looking into this now.

50Wliu:
I am with @Alhadis that adding the `meta.embedded` scope doesn't feel entirely right since we are not embedding the entire JavaScript language.  Is it not possible for VSCode to whitelist the `entity.name.type.instance` scope instead?

mjbvz:
@50Wliu We can use any `meta.embedded.*`, such as `meta.embedded.jsdoctype`. The type is an embedded language in my understanding. How does that sound?

Alhadis:
@mjbvz It's a mistake to treat this as any sort or language, and it's only for the sake of maintainability that JSDoc's patterns have even been separated. Keep in mind this grammar has been stuck with incorrect scope choices for a very long time tgab

Alhadis:
I'm sorry, I pocketed my mobile while typing this, and my phone must have gotten impatient waiting for me to finish... 😓

* Keep in mind that this grammar has been stuck with incorrect scope choices for a very long time, and it's only recently that we've been reaching some sort of consistent and logical structure with relation to our scope choices. Glueing scopes to overcome a technical limitation (not to mention a rather arbitrary one) feels like a step backwards.

mjbvz:
@Alhadis I don't understand why this would be backwards step for the grammar though. In JSDocs, the type is a distinct part of the comment and it uses its own domain specific language. This proposed change seems like a great use case for the meta.embedded scope and I don't think the fact that is driven by our tooling needs makes the proposal less valid

mjbvz:
We added a way to explicitly override the token kind per scope in VS Code: https://github.com/Microsoft/vscode/commit/d74145ed7817fd07335b5308c31a31f0ed503c04

This accomplishes the same thing as using  around the problem so I'll close this PR

Issue #271, opened 2015-11-26
In re-export 'default' is tokenized as 'illegal'


MaximSokolov:
``` js
export { default as alias } from 'module-name'; // ok
export { default as alias }; // ok
export {
  default as alias // not illegal
} from 'module-name';
```

PR #549, opened 2018-01-31
Fix multiline re-export of the default export

kylekthompson:
### Requirements

### Description of the Change

Fixes the highlighting when re-exporting a default export:

```js
export {
  default as alias,
} from 'module';
```

screen shot 2018-01-31 at 12 57 07 pm

### Alternate Designs

Find an alternate regex.

### Benefits

The export will be properly tokenized.

### Possible Drawbacks

There could be a performance hit with the new regex.

### Applicable Issues

I actually did find one! #271

### DIFF
--- a/grammars/javascript.cson
+++ b/grammars/javascript.cson
@@ -177,53 +177,6 @@
         'include': '#numbers'
       }
       {
-        # { member1 , member2 as alias2 , [...] } inside re-export
-        'begin': '{(?=.*\\bfrom\\b)'
-        'beginCaptures':
-          0:
-            'name': 'punctuation.definition.modules.begin.js'
-        'end': '}'
-        'endCaptures':
-          0:
-            'name': 'punctuation.definition.modules.end.js'
-        'patterns': [
-          {
-            # (default|name) as alias
-            'match': '''(?x)
-              (?: \\b(default)\\b | \\b([a-zA-Z_$][\\w$]*)\\b)
-              \\s*
-              (\\b as \\b)
-              \\s*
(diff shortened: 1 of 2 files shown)


kylekthompson:
@50Wliu sorry it's taken me so long to get back to this!

So, I've tracked down the issue with this. The problem is that the multiline re-export is falling into the `from`-less matches [here](https://github.com/atom/language-javascript/blob/master/grammars/javascript.cson#L227-L259).

My proposed change is to search for `from` across multiple lines, though I'm running into some issues with xregexp. Are we not able to use the `s` flag?

MaximSokolov:
> My proposed change is to search for `from` across multiple lines

This is not possible due to limitation of the regex engine. It can be fixed by removing [this pattern]( https://github.com/atom/language-javascript/blob/4342f4705a808c6586ca2651f99cceb623b0a74a/grammars/javascript.cson#L179-L225)  and changing `invalid.illegal` -> `variable.language.default` [here](https://github.com/atom/language-javascript/blob/4342f4705a808c6586ca2651f99cceb623b0a74a/grammars/javascript.cson#L241), so `default` always would be legal inside export statement

kylekthompson:
@MaximSokolov I had considered doing that, but it breaks this case:

```js
export { default as alias };
```

Maybe it's okay that we don't highlight this as invalid?

MaximSokolov:
Yeah, it's fine. Linting is not a priority for this package

kylekthompson:
✨ updated the fix. Thanks for the quick feedback on this!

PR #554, opened 2018-02-12
Add regular expression lookbehind highlighting.

CentricStorm:
### Description of the Change

 1. Added support for regular expression lookbehind syntax highlighting, as the ECMAScript proposal has reached [stage 4](https://github.com/tc39/proposals/blob/master/finished-proposals.md), and is considered a 'finished proposal'.
2. Added unit tests for both lookahead and lookbehind syntax highlighting.

### Alternate Designs

The `meta.assertion.look-behind.regexp` scope was chosen so as to be consistent with the existing `meta.assertion.look-ahead.regexp` scope, though technically `meta.assertion.lookbehind.regexp` would have been better, as the term lookahead is used in the ECMAScript language specification with no dash.

### Benefits

Enables custom syntax highlighting of regular expression lookbehind assertions.

### Applicable Issues

There weren't previously any unit tests testing regular expression lookahead/lookbehind highlighting, so there weren't any to test against.

### DIFF
--- a/grammars/regular expressions (javascript).cson
+++ b/grammars/regular expressions (javascript).cson
@@ -43,14 +43,18 @@
         'name': 'keyword.operator.or.regexp'
       }
       {
-        'begin': '(\\()((\\?=)|(\\?!))'
+        'begin': '(\\()(?:(\\?=)|(\\?!)|(\\?<=)|(\\?<!))'
         'beginCaptures':
           '1':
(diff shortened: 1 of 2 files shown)


50Wliu:
> the ECMAScript proposal has reached stage 4, and is considered a 'finished proposal'.

:tada:

> Added unit tests for both lookahead and lookbehind syntax highlighting

😍

Issue #558, opened 2018-02-15
Syntax inconsistencies using tree-sitter


mikemcbride:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Reproduced the problem in Safe Mode: http://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode
    * Followed all applicable steps in the debugging guide: http://flight-manual.atom.io/hacking-atom/sections/debugging/
    * Checked the FAQs on the message board for common solutions: https://discuss.atom.io/c/faq
    * Checked that your issue isn't already filed: https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom
    * Checked that there is not already an Atom package that provides the described functionality: https://atom.io/packages

### Description

When enabling the experimental `tree-sitter` parsing system, there are some inconsistencies in the syntax highlighting.

### Steps to Reproduce

1. Enable experimental tree-sitter parsers
2. Open a JavaScript file
3. Syntax highlighting is not consistent with old parser

**Expected behavior:** For the most part, I would expect consistency in the syntax highlighting.

**Actual behavior:** There are quite a few inconsistencies between the old and new.

**Reproduces how often:** 100%

### Versions

Atom version: 1.25.0-beta0
APM version: 1.19.0
OS: macOS Sierra 10.12.6

### Additional Information

Here's a screenshot, with the old on the left and the new on the right:

![image](https://user-images.githubusercontent.com/6248612/36240284-48804020-11d6-11e8-90e9-b3809f20764f.png)

My apologies if this is actually expected behavior - I do realize that tree-sitter aims to generate a more accurate syntax tree and that would inevitably cause some highlighting differences. Some of these things do seem to provide extra consistency, but I'm unsure if others are intended behavior or missing features. More specifically, I am curious about the following:

- `const` declarations are no longer highlighted. This one seems like it's probably an improvement in consistency since `let` and `var` are not highlighted orange.
- `module` keyword is no longer highlighted
- `=` sign is no longer highlighted (missing the `syntax--assignment` class in the markup that enables this)
- `require` is highlighted a different color. This one I could also see being an improvement, as function names in this color scheme are also blue, but with `require` being a special case, I wasn't sure.
- the `=>` arrow function symbol is no longer highlighted
- the `this` keyword is no longer highlighted

[...]

PR #562, opened 2018-03-01
Add missing operators

mmcbride1007:
### Description of the Change

This adds additional symbols to the tree-sitter grammar for `keyword.operator.js` to have more consistency with the non-tree-sitter grammar. See comparison screenshots when using tree-sitter vs old grammar:

**OLD:**
image

**TREE SITTER:**
image

### Alternate Designs

It's possible that these were intentionally left out, so we could just leave it the way it is.

### Benefits

All operators would have the same classes on them and would be styled consistently.

### Possible Drawbacks

There may be situations that I'm not thinking of where one of these operators would be used and we would want a different class on it than operator.

### Applicable Issues

Related to #558 

cc: @maxbrunsfeld if you have any suggestions here please let me know!


maxbrunsfeld:
Thanks @mmcbride1007! I wasn't very methodical when I initially created these scope mappings; really glad you're taking the time to improve them.

mmcbride1007:
No worries, happy to help!

PR #570, opened 2018-04-15
remove strange tab stops in DOM selector snippets

Flying-Toast:
### Description of the Change
I removed a strange tab stop in all of the DOM selector snippets (querySelector, getElementById, etc).

For example, the "qs" snippet completes to:
`querySelector('query')` with everything inside the parenthesis as the first tab stop.
I have deleted this tab stop, and now the first tab stop is everything inside the _quotes_ instead.
### Benefits
The DOM selector functions take a string, and I can not think of any practical cases where anything except a string literal would be passed to them, so it doesn't make sense why the quotes were initially part of the first tab stop. As the first tab stop is highlighted/selected when the snippet is inserted, typing after the snippet is inserted would delete the quotes.
### Possible Drawbacks
None

### DIFF
--- a/snippets/language-javascript.cson
+++ b/snippets/language-javascript.cson
@@ -52,22 +52,22 @@
     'body': 'function* ($1) {\n\t$2\n}'
   'getElementsByClassName':
     'prefix': 'get'
-    'body': 'getElementsByClassName(${1:\'${2:className}\'})$3'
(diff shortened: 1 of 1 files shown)


maxbrunsfeld:
I think the motivation for the current tab stops is that you might pass a variable or a template string as the selector, id, class, etc. Because there are valid reasons for the current configuration, I'd like to avoid the churn of changing it.

PR #592, opened 2018-08-26
Colour generator functions too

Aerijo:
### Description of the Change

Generator function names were not given scopes; now they are.

It seems [`language-typescript`](https://github.com/atom/language-typescript/blob/master/grammars/tree-sitter-typescript.cson) could also benefit from this.


maxbrunsfeld:
Thanks!

PR #429, opened 2016-09-04
Add support for gql queries via graphql-tools

tonyxiao:
https://github.com/apollostack/graphql-tools/issues/123

PR #604, opened 2018-10-14
Add support for gql template literal

Ingramz:
Closes #429

cc @50Wliu

### DIFF
--- a/grammars/javascript.cson
+++ b/grammars/javascript.cson
@@ -1429,7 +1429,7 @@
         ]
       }
       {
-        'begin': '(Relay\\.QL)\\s*(`)'
+        'begin': '(Relay\\.QL|gql)\\s*(`)'
         'beginCaptures':
           '1':
             'name': 'entity.name.function.js'
--- a/spec/javascript-spec.coffee
+++ b/spec/javascript-spec.coffee
@@ -802,6 +802,14 @@ describe "JavaScript grammar", ->
       expect(tokens[6]).toEqual value: ' }', scopes: ['source.js', 'string.quoted.template.graphql.js']
       expect(tokens[7]).toEqual value: '`', scopes: ['source.js', 'string.quoted.template.graphql.js', 'punctuation.definition.string.end.js']
 
+  describe "ES6 tagged gql string templates", ->
+    it "tokenizes them as strings", ->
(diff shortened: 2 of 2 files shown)


FluorescentHallucinogen:
@50Wliu @Ingramz It seems it doesn't work. Could you please check?

Ingramz:
@FluorescentHallucinogen for me it works as intended. You'll need to have a third party [graphql package](https://atom.io/packages/language-graphql) installed. And make sure that you don't use the new Tree-Sitter parser as it is not supported there (yet) (uncheck `Settings -> Core -> Use Tree Sitter Parsers`). Then restart editor/reopen file and it should work.

Ingramz:
@FluorescentHallucinogen it looks like inside HTML grammar, someone has repurposed the CoffeeScript grammar for GraphQL which, while it 'works', is not technically correct.

How it looks like when CoffeeScript is used (notice that type declarations are styled differently):
```coffee
query($search: String, $currentPosition: Int) {
  posts(
    where: { OR: [{ title_contains: $search }, { content_contains: $search }] }
    orderBy: updatedAt_DESC
    skip: $currentPosition
    first: 50
  ) {
    id
    updatedAt
    title
    content
    author(where: { isAdmin: true }) {
      id
      name
      isAdmin
    }
  }
}
```

Compare this to GraphQL grammar (Github uses the same community package):
```graphql
query($search: String, $currentPosition: Int) {
  posts(
    where: { OR: [{ title_contains: $search }, { content_contains: $search }] }
    orderBy: updatedAt_DESC
    skip: $currentPosition
    first: 50
  ) {
    id
    updatedAt
    title
    content
    author(where: { isAdmin: true }) {
      id
      name
      isAdmin
    }
  }
}
```

It is certainly possible to swap it out to CofeeScript for the template literals too, but it would be more appropriate to migrate the script tags to leverage the community package as well in order to introduce better uniformity.

The fact that Github uses this grammar, means that this package already has the community's blessing and one should not be afraid having to use that package if they wish to use GraphQL in Atom.

FluorescentHallucinogen:
@Ingramz

Thank you for the detailed explanation!

It wouldn't be better to change CofeeScript grammar to GraphQL grammar in https://github.com/atom/language-html/pull/191/files? Could you please make PR?

FluorescentHallucinogen:
@Ingramz Thanks for the PR. :+1:

Issue #619, opened 2018-11-01
Tree-sitter no scope on `as`


chbk:
### Steps to Reproduce

1. Enable Settings > Core > Use Tree Sitter Parsers
2. Copy-paste:

```javascript
import utils as tools from 'utils'
export { each as forEach }
```

**Expected behavior:**

`as` should be `keyword.control.js`

**Actual behavior:**

`as` is `source.js`

### Versions

Atom    : 1.32.1
Electron: 2.0.9
Chrome  : 61.0.3163.100
Node    : 8.9.3

Issue #617, opened 2018-11-01
Tree-sitter no scope on array brackets


chbk:
### Steps to reproduce

1. Enable Settings > Core > Use Tree Sitter Parsers
2. Copy-paste:

```javascript
let list = [1, 2, 3];
```

**Expected behavior:**

`[`, `]` should be `punctuation.definition.array.begin|end.bracket.square.js`

**Actual behavior:**

`[`, `]` are `source.js`

### Versions

Atom    : 1.32.1
Electron: 2.0.9
Chrome  : 61.0.3163.100
Node    : 8.9.3

Issue #616, opened 2018-11-01
Tree-sitter no scope on semi-colon


chbk:
### Steps to reproduce

1. Enable Settings > Core > Use Tree Sitter Parsers
2. Copy-paste:

```javascript
let num = 3;
```

**Expected behavior:**

`;` should be `punctuation.terminator.statement.semicolon.js`

**Actual behavior:**

`;` is `source.js`

### Versions

Atom    : 1.32.1
Electron: 2.0.9
Chrome  : 61.0.3163.100
Node    : 8.9.3

Issue #566, opened 2018-03-30
Preserve scope name for arrow functions in tree-sitter


chbk:
The [tree-sitter scope](https://github.com/atom/language-javascript/blob/57a8519aae8bdc2d26363c7edeca7045b58103fc/grammars/tree-sitter-javascript.cson#L103) is different from the [textmate scope](https://github.com/atom/language-javascript/blob/57a8519aae8bdc2d26363c7edeca7045b58103fc/grammars/javascript.cson#L1603) for `=>`. Can it be kept the same? Some syntax highlighting packages color arrows differently.

PR #620, opened 2018-11-01
Add more scopes to the tree-sitter grammar

Ben3eeE:
Fixes https://github.com/atom/language-javascript/issues/619
Fixes https://github.com/atom/language-javascript/issues/617
Fixes https://github.com/atom/language-javascript/issues/616
Fixes https://github.com/atom/language-javascript/issues/566

/cc: @maxbrunsfeld

### DIFF
--- a/grammars/tree-sitter-javascript.cson
+++ b/grammars/tree-sitter-javascript.cson
@@ -94,6 +94,7 @@ scopes:
   'method_definition > property_identifier': 'entity.name.function'
   'call_expression > member_expression > property_identifier': 'entity.name.function'
 
+  'this': 'variable.language'
   'super': 'variable.language'
 
   'identifier': [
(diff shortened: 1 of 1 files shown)


Ben3eeE:
@chbk This fixes the issues you reported earlier except `default` is staying scoped as `keyword.control`

chbk:
Great! I tried out your PR and it fixes most issues except:

- `this` is still `source.js`
- `super` and `super()` are scoped differently

Ben3eeE:
Yeah `super()` is a function call and is expected to be scoped differently, while super is a language variable and scoped as such,.

We scope:
`call_expression > super` as `support.function`
and `super` as `variable.language`

`this` is because I messed up when staging. Thanks for noticing and testing 👍 💯 Fixed in the latest commit.

chbk:
Alright, `super()` being scoped differently makes sense. Would it be possible be more specific then, with `support.function.super` for example? It would be nice to be able to highlight `super` the same way in both cases, even if they are syntactically different. After all they refer to the same parent class, `super()` calling its constructor and `super` accessing any of its methods.

Ben3eeE:
I'm ok with making it `support.function.super`. It still looks the same with one-dark-syntax and one-light-syntax.

@maxbrunsfeld I think this is ready for review now 😅

PR #624, opened 2018-11-08
Fold rules for switch_case

Ben3eeE:
### Requirements

### Description of the Change

Folding `case` statements was not included in the tree-sitter grammar. With this PR we add a fold rule that folds up until the `break_statement` or folds the entire case if there isn't a `break_statement`.

All cases folded:
image

Some cases unfolded:
image

### Alternate Designs

Fold everything including the `break_statement`

### Benefits

* You can fold case
* You can see fall through structure very quickly when folded

### Possible Drawbacks

Someone wants the entire case to be folded including the break and we don't allow it!

### Applicable Issues

None

### DIFF
--- a/grammars/tree-sitter-javascript.cson
+++ b/grammars/tree-sitter-javascript.cson
@@ -41,6 +41,15 @@ folds: [
     start: {index: 0, type: '('}
     end: {index: -1, type: ')'}
   }
+  {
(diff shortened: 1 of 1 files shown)


maxbrunsfeld:
This is awesome. Great idea keeping the `break` outside of the fold.

Issue #596, opened 2018-09-16
Js For of loop error in atom


axe-z:
Simple, latest version 1.30

for loop,
the for of autocompletes (without any added snippets) this:

for (variable of iterable) {

}

always f...s me up later.

should be
for (LET variable of iterable) {

}

chfritz:
```
for (i in [1,2,3]) console.log(i)
```
seems to work just fine. Aren't variables that are declared without let/const implicitly `let`s?

axe-z:
It's not a huge deal anyway, but the for IN is ok(var), regular for loop too : 

for (var variable in object) {
  if (object.hasOwnProperty(variable)) {

  }
} && for (var i = 0; i < array.length; i++) {
  array[i]
} 

Might also be time to use the LET keyword ! 
I always, for no good reason mind you... , wake up and figure out the variable's missing after a few seconds of debugging.and YES, I've created my own snippet by now, but still, this should be an easy fix.

fd-rey:
I've the same issue, for...of is creating code that produces ReferenceError.
I think that a possible solution could be modify this file:
https://github.com/atom/language-javascript/blob/master/snippets/language-javascript.cson

setting this line
`'for of':
    'prefix': 'forof'
    'body': 'for (${1:variable} of ${2:iterable}) {\n\t$3\n}'`

to this one (add var before the variable)
`'for of':
    'prefix': 'forof'
    'body': 'for (var ${1:variable} of ${2:iterable}) {\n\t$3\n}'`

axe-z:
left for vscode. Never thought I would.

PR #631, opened 2018-11-21
Fix for...of autocomplete, variable declaration

fd-rey:
### Requirements

### Description of the Change

Following the for...of reference: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of

The variable declaration with "var" is missing and the autogenerated snippet produces a ReferenceError
if the variable used in the for is not declared previously in the code.

### Benefits

The forof autocomplete snippet can be used safely.

### Possible Drawbacks

If the user wants to use an already declared variable, the declaration with "var" in the for has to be removed manually. 

### Applicable Issues

https://github.com/atom/language-javascript/issues/596


jasonrudolph:
Thanks, @fd-rey. :zap:

PR #637, opened 2019-01-23
improve (for) loop

Ali-00-Ridha:
### Requirements

### Description of the Change

fixes #637.
improve for loop structure to improve performance as the new for loop structure reads the array length only one time instead of the old structure which read the array length on every iterate.

new for loop structure:
----
for (var i = array.length; i >= 0; i--) {
	array[i];
}

### Alternate Designs

there is no alternatives

### Benefits

improve for loop structure by reading the array length only one time instead of reading array length on every iterate.

### Possible Drawbacks

maybe it's the new structure has harder readability than the old structure.

### Applicable Issues

improve loop performance

### DIFF
--- a/snippets/language-javascript.cson
+++ b/snippets/language-javascript.cson
@@ -28,7 +28,7 @@
     'body': 'else if (${1:true}) {\n\t$2\n}'
   'for':
     'prefix' : 'for'
-    'body' : 'for (var ${2:i} = 0; ${2:i} < ${1:array}.length; ${2:i}++) {\n\t${1:array}[${2:i}]$3\n}'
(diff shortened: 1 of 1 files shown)


Arcanemagus:
Thank you very much for the contribution!

We generally don't accept modifications for snippets in the language packages since they are easily overridden and we expect that people will do so for their own purposes. You can find instructions on how to create your own snippets in the [Flight Manual](https://flight-manual.atom.io/using-atom/sections/snippets/). Viewing a language package that provides snippets allows you to copy the current version as a starting place for creating your own.

Thanks for the suggestion but we're going to leave things as they are for now unless the current snippet produces invalid code.

PR #638, opened 2019-01-25
Add CSS, GraphQL, and SQL template literal tags to Tree-Sitter grammar

bennypowers:
### Description of the Change

Similar to the `html`, `gql`, or `sql` template literal tags which are already supported, this PR adds support for a `css` template literal tag.

My only concern with the PR is with my use of `string.quoted.template.css.js` on line 1384. Does this need to be defined elsewhere?

### Alternate Designs

I think the feature is pretty straightforward, the alternative is not to have the CSS grammar in such templates.

### Benefits

Use cases like the following will receive CSS grammars:

```js
import { LitElement, css } from 'lit-element';
class MyEl extends LitElement {
  static get styles()  {
    return css`
      :host {
        display: block;
      }
    `;
  }
}
```

### Possible Drawbacks

Added maintenance burden.

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -63,10 +63,10 @@ exports.activate = function () {
   }
 }
 
-const STYLED_REGEX = /\bstyled\b/i
+const CSS_REGEX = /\bstyled\b|\bcss\b/i
 
-function languageStringForTemplateTag (tag) {
-  if (STYLED_REGEX.test(tag)) {
+function languageStringForTemplateTag(tag) {
+  if (CSS_REGEX.test(tag)) {
     return 'CSS'
   } else {
     return tag
--- a/lib/main.js
+++ b/lib/main.js
@@ -64,10 +64,16 @@ exports.activate = function () {
 }
 
 const CSS_REGEX = /\bstyled\b|\bcss\b/i
+const GQL_REGEX = /\bgraphql\b|\bgql\b/i
+const SQL_REGEX = /\bsql\b/i
 
 function languageStringForTemplateTag (tag) {
   if (CSS_REGEX.test(tag)) {
     return 'CSS'
+  } else if (GQL_REGEX.test(tag)) {
+    return 'GraphQL'
+  } else if (SQL_REGEX.test(tag)) {
+    return 'SQL'
   } else {
     return tag
   }
--- a/lib/main.js
+++ b/lib/main.js
@@ -65,7 +65,7 @@ exports.activate = function () {
 
 const CSS_REGEX = /\bstyled\b|\bcss\b/i
 
-function languageStringForTemplateTag(tag) {
+function languageStringForTemplateTag (tag) {
   if (CSS_REGEX.test(tag)) {
     return 'CSS'
   } else {
--- a/grammars/javascript.cson
+++ b/grammars/javascript.cson
@@ -1371,6 +1371,30 @@
         ]
       }
       {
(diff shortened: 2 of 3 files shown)


bennypowers:
So, I think the tests are well-formed, but I get results like:


```
Expected { value : '${display', scopes : [ 'source.js', 'string.quoted.template.css.js', 'meta.property-list.css', 'meta.property-value.css' ] } to equal { value : 'display', scopes : [ 'source.js', 'string.quoted.template.css.js', 'source.js.embedded.source' ] }.
```

Note the `value: ${display` which indicates to me that this token isn't being picked up as a js interpolation, despite line 1390: `'include': '#interpolated_js'`

Not sure how to proceed from here.

bennypowers:
It's worth noting that this works out of the box when switching grammars to typescript.

In fact, if I switch grammars to typescript, then switch back to javascript, I still have highlighting in my template.

daviwil:
Hi @bennypowers, since we've moved to using Tree-Sitter for `language-javascript` we would prefer it if you added your change to that grammar configuration instead.  Turns out that this is very easy, you just need to add a check for `'css'` in [this function's](https://github.com/atom/language-javascript/blob/master/lib/main.js#L70) `if` statement and return `CSS`.  No tests are required as we aren't currently testing Tree-Sitter grammar mappings in this repo.

If you're interested in updating your PR to take this approach, I'll be happy to merge it and include it in the next update.  Thanks!

bennypowers:
K cool

If this is good, I'm happy to rebase

lib/main.js
```diff
@@ -63,10 +63,10 @@ exports.activate = function () {
   }
 }
 
-const STYLED_REGEX = /\bstyled\b/i
+const CSS_REGEX = /\bstyled\b|\bcss\b/i
 
-function languageStringForTemplateTag (tag) {
-  if (STYLED_REGEX.test(tag)) {
+function languageStringForTemplateTag(tag) {
+  if (CSS_REGEX.test(tag.toLowerCase())) {
```
daviwil: Shouldn't need the toLowerCase since the regex is already case-insensitive. Otherwise, this looks good!

bennypowers:
should we include `gql` and `sql` here?

daviwil:
Sure, just update the PR title to include those so that it's clear when they got added.

lib/main.js
```diff
@@ -63,10 +63,10 @@ exports.activate = function () {
   }
 }
 
-const STYLED_REGEX = /\bstyled\b/i
+const CSS_REGEX = /\bstyled\b|\bcss\b/i
 
-function languageStringForTemplateTag (tag) {
-  if (STYLED_REGEX.test(tag)) {
+function languageStringForTemplateTag(tag) {
```
daviwil: At the risk of sounding nitpicky, can you add back the space before the open parentheses? That's part of the code style across Atom repos. Thanks!

Aerijo: @daviwil I thought all repos had linters as part of the tests?

daviwil: I only see `coffeelint` hooked up on this repo, must not have had `standard` added to it once JavaScript code was introduced.

bennypowers: no problemo

bennypowers: atom-ide-ui has a bad habit of applying formatting on save, regardless of eslint

bennypowers:
Please check language string on line 74

daviwil:
Looks great, thanks a lot @bennypowers!

Issue #641, opened 2019-02-27
Remove invalid string highlighting


50Wliu:
From https://github.com/github/linguist/issues/3044#issuecomment-462016949

PR #642, opened 2019-03-15
Remove invalid string highlighting from TextMate grammar

lildude:
### Requirements

### Description of the Change

GitHub's [Linguist](https://github.com/github/linguist), which is used for syntax highlighting on GitHub.com, currently uses this grammar for highlighting Javascript and JSX. For quite some time now, users have experienced the problem where the presence of an apostrophe within JSX causes the syntax to be highlighted in red as an error:

 

@Alhadis [proposed](https://github.com/github/linguist/issues/3044#issuecomment-459651756) the solution of removing the "rule which highlights unterminated strings as errors" which @50Wliu suggested would probably be fine as the TextMate grammars are no longer actively maintained:

> @Alhadis that's probably fine. We don't actively maintain the TextMate variants anymore, but I think this is something we'll merge a PR for.

@50Wliu opened https://github.com/atom/language-javascript/issues/641 to track this. I'm now implementing the changes suggested by @Alhadis in this PR.

This change _only_ affects the TextMate grammar as this is the grammar Linguist uses as tree-sitter grammars are not currently supported.

### Benefits

Syntax highlighting on GitHub.com for Javascript with JSX that contains apostrophes will be rendered correctly.

### Possible Drawbacks

As this change only affects the TextMate grammar, there is a risk that the editor behaviour may vary slightly from the representation on GitHub.com for as long as we use the TextMate grammar.

### Applicable Issues

See https://github.com/github/linguist/issues/3044 for further details of the problem we're trying to resolve.

Fixes https://github.com/atom/language-javascript/issues/641

### DIFF
--- a/grammars/javascript.cson
+++ b/grammars/javascript.cson
@@ -1344,10 +1344,6 @@
           {
             'include': '#string_escapes'
           }
-          {
-            'match': "[^']*[^\\n\\r'\\\\]$"
-            'name': 'invalid.illegal.string.js'
-          }
         ]
       }
       {
@@ -1363,10 +1359,6 @@
         'patterns': [
           {
             'include': '#string_escapes'
-          }
-          {
-            'match': '[^"]*[^\\n\\r"\\\\]$'
-            'name': 'invalid.illegal.string.js'
           }
         ]
       }
--- a/spec/javascript-spec.coffee
+++ b/spec/javascript-spec.coffee
@@ -51,7 +51,6 @@ describe "JavaScript grammar", ->
           line3
         """ + delim
         expect(lines[0][0]).toEqual value: delim, scopes: ['source.js', scope, 'punctuation.definition.string.begin.js']
-        expect(lines[0][1]).toEqual value: 'line1', scopes: ['source.js', scope, 'invalid.illegal.string.js']
         expect(lines[1][0]).toEqual value: 'line2\\', scopes: ['source.js', scope]
         expect(lines[2][0]).toEqual value: 'line3', scopes: ['source.js', scope]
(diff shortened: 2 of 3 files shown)


lildude:
🤔 Sorting out the tests is going to be interesting as we only want these changes to be considered when testing the TextMate grammar. I'm open to suggestions on how best to do this.

lildude:
Ooo, maybe not. If I'm reading the tests correctly, the tests only run against the TextMate grammar:

https://github.com/atom/language-javascript/blob/e5034e28f00e5401890516b7bce47fe3a9041c3c/spec/javascript-spec.coffee#L14-L15

So whatever test changes I made, they'll only apply to the testing with the TextMate grammar.

50Wliu:
I thought about this for a bit: wouldn't this still make it so everything will be matched as a string until the next quote? I suppose that's better than angry red highlighting, but still not that ideal.

lildude:
> I thought about this for a bit: wouldn't this still make it so everything will be matched as a string until the next quote?

Yup. I've gone back to @Alhadis to see if he's got a better idea or implementation method as I agree, it's not ideal, however it's certainly waaaay better than all the red.

lildude:
Heard back:

> I thought about this for a bit: wouldn't this still make it so everything will be matched as a string until the next quote?

Yup, but this would only affect GitHub and Atom users that have the treesitter grammars disabled. An alternate solution is waaay too complex and requires changes to Atom.

50Wliu:
All right, let's do this then.

50Wliu:
🚀 as language-javascript@0.129.21.

PR #643, opened 2019-03-20
Add more prototype methods to support.function scope

atomiks:
### Description of the Change

Adds more prorotype methods to `support.function`. For example with the Atom One Dark theme, these functions are a more cyan color compared to regular functions which are blue.

Includes 

- Array#{map,filter,reduce,reduceRight,flat,flatMap,every,some,includes,copyWithin}
- Function#bind
- Object#hasOwnProperty

### Benefits

Less inconsistency in coloring when using methods

### Possible Drawbacks

None that I am aware of

### Applicable Issues

---

Ref: https://github.com/Microsoft/TypeScript-TmLanguage/issues/699

### DIFF
--- a/grammars/javascript.cson
+++ b/grammars/javascript.cson
@@ -1667,14 +1667,16 @@
               {
                 'match': '''(?x)
                   \\b(catch|finally|then|shift|showModelessDialog|showModalDialog|showHelp|scroll|scrollX|scrollByPages|
-                  scrollByLines|scrollY|scrollTo|stop|strike|sizeToContent|sidebar|signText|sort|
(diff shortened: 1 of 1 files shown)


50Wliu:
Thanks very much for reaching out!

As you may have heard, we are migrating from our old [first-mate grammar engine](https://github.com/atom/first-mate) to the [new Tree-sitter engine](https://blog.github.com/2018-10-31-atoms-new-parsing-system/). This will enable a number of new features, more consistent syntax highlighting, and better performance, among other benefits. In order to free up our limited resources, we have decided to stop maintaining the first-mate grammar when there is a built-in Tree-sitter grammar available. Since this PR applies only to the first-mate grammar, we are going to close it.

Thanks again for your interest in Atom.

PR #645, opened 2019-05-17
Add support for multi-line types in jsdoc. (updated)

duckbrain:
### Requirements

### Description of the Change

This is fork of #515, changing the style as requested by @50Wliu. Note, the other requested change on that MR appears to have already been done by @ribrdb.

Support for multi line types in jsdoc.  
See microsoft/TypeScript-TmLanguage#467

And here's an example file with multi-line types:
https://github.com/google/closure-library/blob/master/closure/goog/dom/animationframe/animationframe.js#L61

### Alternate Designs

I don't know of any.

### Benefits

Multi line types are correctly highlighted.

### Possible Drawbacks

Unterminated types are not recognized as invalid. I couldn't figure out any way to maintain that. However you can clearly see that the type runs on longer than you're expecting if you forget the closing }, so I don't think that's a major issue.


rsese:
Thanks @duckbrain!  I'm not sure if you've seen, but since https://github.com/atom/language-javascript/pull/515 was initially opened, we've been migrating from our old [first-mate grammar engine](https://github.com/atom/first-mate) to the [new Tree-sitter engine](https://blog.github.com/2018-10-31-atoms-new-parsing-system/). This will enable a number of new features, more consistent syntax highlighting, and better performance, among other benefits. In order to free up our limited resources, we have decided to stop maintaining the first-mate grammar when there is a built-in Tree-sitter grammar available as is this case here with JavaScript.

Since this PR applies only to the first-mate grammar we're going to close this out but please let us know if we've misunderstood anything and thanks again for your contribution.

SamB:
@rsese: So the [TypeScript-TmLanguage](https://github.com/microsoft/TypeScript-TmLanguage) folks should *probably* stop sending people with changes for the jsdoc grammar your way.

I guess I'll go suggest an additional comment on the relevant portion of [TypeScript.YAML-tmLanguage](https://github.com/microsoft/TypeScript-TmLanguage/blob/master/TypeScript.YAML-tmLanguage)...

rsese:
@SamB - yes, we're not prioritizing issues across language repositories that are TextMate specific when a Tree-sitter grammar exists for that language.

Issue #560, opened 2018-02-20
Apostrophe in comment reduces indentation level on auto-indent


wbt:
Both of the following were auto-indented by Atom.   
This code looks properly indented:

    function example(param1) { //you wouldnt expect this apostrophe to affect indentation!
        var starter = doSomeFormatting(param1,
            {message: "hardcoded",
            information: extra,
            more: beyond}
        );
        var a = 4;
        var b = 5;
        return add(a,b);
    }

This code does not:

    function example(param1) { //you wouldn't expect this apostrophe to affect indentation!
    var a = 4;
    var b = 5;
    var starter = doSomeFormatting(param1,
        {message: "hardcoded",
        information: extra,
        more: beyond}
        //The negative indentation error suggest that additional code here is no longer
        //a sub-block within the function example().
        //If there are multiple such comments, the indentation errors accumulate quickly.
    );

and the only difference between them is a commented-out apostrophe, on a line after an open brace!

I noticed the [same issue in Solidity](https://github.com/jesseweed/language-solidity/issues/1) and the issue may generalize beyond this package, but my [last](https://github.com/atom/atom/issues/16720) [attempts](https://github.com/atom/atom/issues/16721) at opening auto-indent issues on Atom got moved to [language](https://github.com/atom/language-javascript/issues/552)-[specific](https://github.com/atom/language-javascript/issues/553) packages (so I'm starting out there with these reports).  

### Versions

Atom: 1.23.2
Electron: 1.6.15
Chrome: 56.0.2924.87
Node: 7.4.0
language-ethereum: 0.3.9
language-javascript: 0.127.6
OS : Windows 10

PR #650, opened 2019-06-25
Fix auto-indent when the previous line ends with a comment with a quote

### LINKED ISSUE #648: Quote character in comment causes auto-indent error
### Description
When a comment contains a quote character (`'` or `"`) on a line which should be followed by an additional level of indentation, all following lines are indented one level less than they should be.

### Steps to Reproduce
1. In Atom, create a new Javascript file.  (Optionally configure 1 indent = 4 spaces). 
2. Write or paste code like the following:
```
if(firstBoolean) {
    var myVar1 = "Indentation from the first conditional makes the issue more visible.";
    if(secondBoolean) { //this comment has at least one "double-quote" character in it
        var myVar2 = "This content and everything that follows is one indent level left of correct.";
        return myvar2;
    }
}
```
3. Select All.
4. Auto-indent.  

**Expected behavior:** 
Code formatted like the sample above.
Note that the close curly brace for the `if (secondBoolean)` is aligned with the if statement, and the contents of that conditional block are indented one level farther right. 

**Actual behavior:** 
```
if(firstBoolean) {
    var myVar1 = "Indentation from the first conditional makes the issue more visible.";
    if(secondBoolean) { //this comment has at least one "double-quote" character in it
    var myVar2 = "This content and everything that follows is one indent level left of correct.";
    return myvar2;
}
}
```
Note that everything after `if(secondBoolean)` is indented one level less than it should be, until the point when there is no further left the code can go.

Removing all quote characters and doing another auto-indent causes proper alignment.  It does not appear to matter where in the comment the quote character appears.  Even a quote character in a contraction will trigger the bug. 

**Reproduces how often:**
Always, even in safe mode. 

### Versions
Atom    : 1.23.2
Electron: 1.6.15
Chrome  : 56.0.2924.87
Node    : 7.4.0
Windows: 10 Pro

as-cii:
Fixes #648

🍐'd with @nathansobo


wbt:
Does this also fix [#560](https://github.com/atom/language-javascript/issues/560)?  (I have not tested.)

PR #653, opened 2019-07-16
Fix lost colorization of function declaration names

caleb531:
### Requirements

### Description of the Change

A recent update to https://github.com/tree-sitter/tree-sitter-javascript restructured the way function declarations are constructed in the syntax tree. This change is not present in the latest stable Atom (v1.38.0 at the time of this writing), but is present when I forked language-javascript (v0.132.0 at the time of this writing).

In effect, when I cloned and linked `language-javascript`, the names for function declarations are no longer colorized.

**Current:**  
1-before

This PR restores colorization for function declaration names, according to the new syntax tree structure.

**After this PR is merged:**  
2-after

**Syntax tree in Atom 1.38.0 (working):**  
3-tree-old-working

**Syntax tree in language-javascript 0.132.0 (what breaks things):**  
4-tree-new-broken

### Alternate Designs

A different selector could theoretically be used, but `function_declaration > identifier` seemed most appropriate and consistent with what was already present in the grammar.

### Benefits

Colorization for function declaration names is restored to its former glory!

### Possible Drawbacks

None that I can see, as the change is small and limited in scope, and per my screenshot above, function expressions are not negatively impacted by this change.

### Applicable Issues

N/A

### DIFF
--- a/grammars/tree-sitter-javascript.cson
+++ b/grammars/tree-sitter-javascript.cson
@@ -97,6 +97,7 @@ scopes:
   'call_expression > identifier': {match: '^[A-Z]', scopes: 'meta.class'}
 
   'function > identifier': 'entity.name.function'
+  'function_declaration > identifier': 'entity.name.function'
   'generator_function > identifier': 'entity.name.function'
 
   'call_expression > identifier': [


rsese:

jasonrudolph:
Nice catch, @caleb531. Thanks for this fix! :zap:

PR #654, opened 2019-07-17
Colorize rest parameters like other formal parameters

caleb531:
### Requirements

### Description of the Change

Currently, the tree-sitter grammar does not colorize rest parameters in functions. This is in contrast to the first-mate grammar, which _does_ colorize such parameters. This PR colorizes rest parameters as `formal-parameter.identifier`, like other function parameters.

**Before:**  
1-before

**After (with these PR changes):**  
2-after

**Current first-mate grammar:**  
3-firstmate

### Alternate Designs

The selector was chosen because it is similar to the existing selector for formal parameters, so it should have no conflicts with other grammar rules.

### Benefits

Rest parameters are now treated just like other formal function parameters.

### Possible Drawbacks

None that I can see, since the grammar rule's selector is limited in scope.

### Applicable Issues

N/A

### DIFF
--- a/grammars/tree-sitter-javascript.cson
+++ b/grammars/tree-sitter-javascript.cson
@@ -70,7 +70,7 @@ scopes:
   'member_expression > property_identifier': 'variable.other.object.property.unquoted'
 
   'formal_parameters > identifier': 'variable.parameter.function'
-  'formal_parameters > rest_parameter > identifier': 'variable.parameter.function'
+  'formal_parameters > rest_parameter > identifier': 'variable.parameter.rest.function'
 
   'shorthand_property_identifier': [
     {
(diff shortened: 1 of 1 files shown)


rsese:

jasonrudolph:
@caleb531: Similar to https://github.com/atom/language-javascript/pull/652#issuecomment-521430421, I think we may want to use the scope names that the TextMate grammar uses. What do you think?

### Scopes assigned using TextMate

Screen Shot 2019-08-16 at 9 08 04 AM


### Scopes assigned using Tree-sitter grammar on this branch

Screen Shot 2019-08-16 at 9 07 44 AM

caleb531:
@jasonrudolph Done! The scope is now `variable.parameter.function`.

jasonrudolph:
Thanks for this, @caleb531! :zap::bow:

PR #664, opened 2019-10-31
Add a 'forEach'  snippet

pbsds:
### Requirements

### Description of the Change

It is the simple addition of a single snippet. I did not find tests for these in previously merged commits of the same nature.

### Alternate Designs

I could instead use an anonymous function instead of arrow notation, but after a quick check on `caniuse.com` i decided against it.

### Benefits

sanity++

### Possible Drawbacks

Not all browsers may support arrow notation


darangi:
@pbsds, This is awesome! ❤️

PR #681, opened 2020-10-18
globalThis, self, frames

icecream17:
### Description of the Change
Adds:
```
globalThis in 'support.variable'
self and frames in 'support.variable.dom'
```

`self === window`, and `frames === window`

Also, globalThis was added to js. It's a way to reference the value of (this) in any global scope.

MDN links:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis
https://developer.mozilla.org/en-US/docs/Web/API/Window/window
https://developer.mozilla.org/en-US/docs/Web/API/Window/self
https://developer.mozilla.org/en-US/docs/Web/API/Window/frames

Currently only updates tree-sitter-javascript.cson, and not javascript.cson

### Benefits
This is only a minor update

### Possible Drawbacks
globalThis doesn't always point to a DOM object, so it's only in support.variable

In a worker context, self resolves to `WorkerGlobalScope.self` instead of `window.self`, so maybe also put that in support.variable

### DIFF
--- a/grammars/javascript.cson
+++ b/grammars/javascript.cson
@@ -1,3 +1,6 @@
+# Note: Because of the optional chaining operator (?.), "object.property" in the comments usually means ("object.property" or "object?.property")
+# Even though sometimes it doesn't seem like using the chaining operator would make any sense, it's still supported. example: Math?.PI
+# Negative lookbehinds for "." Shouldddd cover "?."
 'scopeName': 'source.js'
 'fileTypes': [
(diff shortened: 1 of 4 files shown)


icecream17:
Today I did 4 more commits, but they're not supposed to be part of the marge request

icecream17:
I'm making a new pull request with a branch - see previous comment

PR #683, opened 2020-10-29
Use let instead of var in the snippets

aminya:
### Description of the Change

Use `let` instead of `var` in the snippets.

### Benefits
`var` is not recommended, and it is detected as an error in most of the modern linters.


darangi:
Hey @aminya Thanks for contributing, However, we are not ready to accept this PR at the moment

PR #684, opened 2020-11-12
GitHub Actions

aminya:
### Description of the change

Use GitHub Actions instead of Travis and Appveyor. 

### Benefits
- GitHub Actions is much faster than the other CI services and allows parallel runs 
- GitHub Actions it is much easier to configure. 
- Atom is a GitHub product, so we should use GitHub Actions.
- Travis is making its services limited and monetary.   
https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing

### Verification 
The CI passes: https://github.com/aminya/language-javascript/actions/runs/359738214

### Drawbacks
none

### Release Notes
N/A

Includes and closes #102


icecream17:
(Probably a silly question)
Is AppVeyor removed because it's redundant?

aminya:
Appveyor was being used for testing the code on Windows. But now GitHub Actions has Windows support, we don't need Appveyor anymore.

P.S: Because I don't have access to this organization, I cannot set the "ignore build if no appveyor.yml" option, which disables Appveyor when appveyor.yml is deleted. So I have to use a dummy appveyor.yml to achieve the same thing.

PR #686, opened 2020-11-13
Operators

### LINKED ISSUE #640: Optional chaining not supported in embedded script
### Prerequisites

### Description

Optional chaining is currently supported in Javascript files, but not in embedded script (`<script>` tag in HTML). It is currently treated as a ternary operator which breaks the detection of the closing tag `</script>` since a colon is expected.

I first reported the issue to `language-vue` to finally realized it was the case for all embedded scripts:
https://github.com/hedefalk/atom-vue/issues/97

51517628-a99b7a80-1e1b-11e9-8465-c8dec87ee902

### Steps to Reproduce

1. Open a HTML file
2. Within a `<script>` tag, use the optional chaining operator (i.e. `window.test?.something`)

**Expected behavior:** Closing script tag detected

**Actual behavior:** Closing script tag not detected

**Reproduces how often:** Always

### Versions

Atom: 1.32.2
OS: macOS Mojave version 10.14.3 (18D109)

### LINKED ISSUE #680: Nullish Coalescing Operator (??)
### Prerequisites

    * Reproduced the problem in Safe Mode: <https://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode>
    * Followed all applicable steps in the debugging guide: <https://flight-manual.atom.io/hacking-atom/sections/debugging/>
    * Checked the FAQs on the message board for common solutions: <https://discuss.atom.io/c/faq>
        * Replacing all `??` with `||` does not suffice, and causes errors
    * Checked that your issue isn't already filed: <https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom>
    * Checked that there is not already an Atom package that provides the described functionality: <https://atom.io/packages>
        * Close, but does weird highlighting on large files: <https://atom.io/packages/language-javascript-jsx>

### Description

Atom does not take `??` into account when highlighting/folding

### Steps to Reproduce

1. Use the JavaScript syntax highlighter
2. `Copy/Paste`
```javascript
// ...
// Notifications
NOTIFICATIONS = $('[data-test-selector="onsite-notifications-toast-manager"i] [data-test-selector^="onsite-notification-toast"i]', true).map(
    element => {
        let streamer = {
            live: true,
            href: $('a', false, element)?.href,
            icon: $('figure img', false, element)?.src,
            name: $('figure img', false, element)?.alt,
        };

        if(!defined(streamer.name))
            return;

        element.setAttribute('draggable', true);
        element.setAttribute('twitch-tools-streamer-data', JSON.stringify(streamer));
        element.ondragstart ??= event => {
            let { currentTarget } = event;

            event.dataTransfer.setData('application/twitch-tools-streamer', currentTarget.getAttribute('twitch-tools-streamer-data'));
            event.dataTransfer.dropEffect = 'move';
        };
    }
);
// ...
```
3. Attempt to fold *line 17* (it should be matched with *line 22* but isn't)

### Expected Behavior

1. The JavaScript syntax highlighter should distinguish between `??` ([Nullish Coalescing Operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator)) and `? ...: ...`
2. The highlighter should also allow the correct code-folding for `??`

### Actual Behavior

1. The syntax highlighter ignores `??` and treats it as `? ...: ...`
2. Code-folding currently mismatches `{}` `[]` and `()` due to it thinking `??` &rArr; `? ...: ...`

**Reproduces how often:**

`100%`

### Versions

* Electron Version: `5.0.13`
* Chrome Version: `73.0.3683.121`
* Node Version: `v12.0.0`
* Operating System:  `Windows 10`
* Last known working version: **N/A**

### Screenshots

> ### Incorrect pair being matched (*line 17* &rarr; *line 23*)
> image

> ### Correct pair being matched (*line 17* &larr; *line 22*)
> image

### LINKED ISSUE #715: Oddity with optional chaining
### Prerequisites

    * Reproduced the problem in Safe Mode: <https://flight-manual.atom.io/hacking-atom/sections/debugging/#using-safe-mode>
    * Followed all applicable steps in the debugging guide: <https://flight-manual.atom.io/hacking-atom/sections/debugging/>
    * Checked the FAQs on the message board for common solutions: <https://discuss.atom.io/c/faq>
    * Checked that your issue isn't already filed: <https://github.com/issues?utf8=✓&q=is%3Aissue+user%3Aatom>
    * Checked that there is not already an Atom package that provides the described functionality: <https://atom.io/packages>

### Steps to Reproduce

1. Add this code to the editor:

```js
iterateJsdoc(() => {
  const checkNativeTypes = () => {
    for (const strictNativeType of strictNativeTypes) {
      if (
        !preferredTypes?.[typeNodeName]
      ) {
        continue;
      }
    }

    return changedPreferred;
  };
}, {
  meta: {},
});
```

2. Click to the end of the line with the `for`
3. Use the bracket matcher command (ctrl-m by default)

**Expected behavior:**

The cursor should go to the end of the `for` tloop

**Actual behavior:**

It goes to the beginning of the document in safe mode (but to the end of the document otherwise for some reason), neither of which is expected.

**Reproduces how often:**

100%.

### Versions

```
Atom    : 1.60.0
Electron: 9.4.4
Chrome  : 83.0.4103.122
Node    : 12.14.1
```

```
apm  2.6.2
npm  6.14.13
node 12.14.1 x64
atom 1.60.0
python 3.9.7
git 2.32.0
```

OS: Mac OS 12.3.

### Additional Information

Changing the `return` line to `return true` gives yet another wrong behavior.

If I get rid of the optional chaining or put the parentheses on the same line, or get rid of the `meta` object, it works as expected.

icecream17:
- <s>Previously blocked by https://github.com/tree-sitter/tree-sitter-javascript/pull/152</s> (and other updating prs)
- <s>Previously blocked by #692</s>
    - Blocked until https://github.com/atom/atom/issues/22129 is fixed
    - Well, it works in legacy tree-mate, but not in tree-sitter.
    - Help is very appreciated

# Description of the Change
Edit: update

Adds optional chaining `a?.b` `a?.[5]` `a?.()`
Adds `**=`
Adds logical assignment `&&= ||= ??=`
Add `shift` class to `>>= >>>= <<=`
Add `**`

Ugh unfinished edit this later

### Alternate Designs
<s>I don't see how to make the ?. and . distinct without making everything that uses it seperate:

It's much bigger than this but the idea is the same:

Original:
```"var.prop" -> "variable" "meta.delimiter.period" "property"```

Two way I thought of to support ?.
```"var(?. or .)prop" -> "var" "meta.delimiter" "property"```

or
```
"var.prop" -> "var" "meta.delimiter.period" "property"
"var?.prop" -> "var" "meta.delimiter.optional" "property"
```

I still added ?. as "meta.delimiter.optional" in the tree-sitter, but in the non-tree-sitter, they're the same.</s>
Edit: They're separate selectors, there's [a later update below](https://github.com/atom/language-javascript/pull/686#issuecomment-770310223).

I've added tests for the ?. operator, and included ?? in the logic operator test. (Works!!!!!)

### Benefits
Fixes #680
Fixes #640
Fixes #715 (duplicate)

### Possible Drawbacks
A lot of comments
<s>There might be a way to make them separate without adding a bunch of redundant code, but I haven't figured it out yet.</s> I actually did figure it out, again there's an update below

### Applicable Issues
#640 #680

### CODE REVIEW

**Comment by icecream17** [comment]
Only seems to fix ?? right now, so the current code doesn't do what I thought (of course...)
image

It thinks the ? is ternary, and the period to be a.. regular period.

**Comment by icecream17** [comment]
Oh the latest issue made me realize I didn't even put in
hmm?.[2]
and
hmm.jump?.()

I think I should reset the files to atom:master to retry fixing this

**Comment by utkarshgupta137** [comment]
Shouldn't we also include "??=" in operators?

**Comment by icecream17** [comment]
I added the missing operators but I broke something and the tests are failing

**Comment by darangi** [comment]
Hi @icecream17 could you take a look into the failing tests?

**Comment by icecream17** [comment]
Current status
exampleFunc?.() not supported
exampleArr[7]?.() not supported
a?.b should be support if it works
a?.b() should also be supported if it works

> 
> 
> Shouldn't we also include "??=" in operators?

Oh yeah. That's added
As for the errors, I try to figure them out tomorrow - I don't know what's wrong, except that regular unrelated code isnt parsing so there's some catastrophic error somewhere

**Comment by icecream17** [comment]
359 fails, now down to 66, though https://github.com/atom/language-javascript/pull/686/commits/6f32f5ad84fe1ca2034c104a27e34c39a4ec7c3a has only 13 errors, so maybe I should just use that instead....

maybe I should provide less updates?

**Comment by icecream17** [comment]
It passed!!!!!!

Summary of changes:
image

Non tree-sitter scope changes:
```
optional matches "?."

meta.delimiter.method.js -->
   meta.delimiter.method.period.js
   meta.delimiter.method.optional.js
meta.delimiter.property.js -->
   meta.delimiter.property.period.js
   meta.delimiter.property.optional.js

keyword.operator.assignment.compound.logical.js added
   used by &&= ||= ??=
```

**Comment by icecream17** [comment]
Oh wait, it only works in the Non-Tree Sitter Parser

File: grammars/javascript.cson [hunk 1]
```diff
@@ -1,3 +1,6 @@
+# Note: Because of the optional chaining operator (?.), "object.property" in the comments usually means ("object.property" or "object?.property")
+# Even though sometimes it doesn't seem like using the chaining operator would make any sense, it's still supported. example: Math?.PI
+# Negative lookbehinds for "." Shouldddd cover "?."
```
sadick254, line 3:
```suggestion
# Negative look-behinds for "." Should cover "?."
```

icecream17:
There's some precedence thing with the operator parsing. I'm not sure how it works.

I found the actual parser: https://github.com/tree-sitter/tree-sitter-javascript
And the grammar file: https://github.com/tree-sitter/tree-sitter-javascript/blob/master/src/grammar.json
_______
<sup>Note to self: https://github.com/atom/language-javascript/blob/master/grammars/tree-sitter-javascript.cson just maps parsed nodes to names.<sup>

icecream17:
image

The tree-sitter parser atom uses can't actually parse some operators yet. Will update when it can.

icecream17:
Update: 
See https://github.com/atom/atom/issues/22129
Because of that error it's blocked until atom is updated: https://github.com/atom/atom/pull/22130

icecream17:
Ok, even though it still doesn't work in tree-sitter, it could still be merged as long as there's no bad changes.

_____
### Incompatible changes

Currently I've "added scopes" in https://github.com/atom/language-javascript/pull/686/commits/743fcd5b42fe9efe1f91c61ea7c9aee0c089f946 but https://github.com/atom/language-javascript/pull/690/files takes care of that.

PR #691, opened 2021-02-10
use let/const es6 syntax for, forin, forof

mrbbp:
### Requirements

### Description of the Change
sorry first time,

i use the `for`, `forin` and `for...of` snippets but my linter configured to works with es6 is unhappy.
so i propose the change `var` to more contemporary variable declaration `let` and `const`

### Benefits

more contemporary variable declaration

### Possible Drawbacks
 possibly confused for beginners

### Applicable Issues
perhaps a prob with old navigator

### DIFF
--- a/snippets/language-javascript.cson
+++ b/snippets/language-javascript.cson
@@ -28,13 +28,13 @@
     'body': 'else if (${1:true}) {\n\t$2\n}'
   'for':
     'prefix' : 'for'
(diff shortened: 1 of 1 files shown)


apicgg:
This makes sense. Thanks. @sadick254 Could you please merge this to master? It will be really helpful to have 'let' instead of 'var'.

icecream17 (approved):

Issue #696, opened 2021-07-14
Windows github actions not working


icecream17:
I copied the branch `operators` into a new branch `non-tree-sitter-operators`

Awhile ago, https://github.com/atom/language-javascript/pull/686/checks?check_run_id=2264799413 passed
But now it fails: https://github.com/icecream17/language-javascript/runs/3069847194

Meaning that at one point the windows github actions stopped working.

PR #698, opened 2021-07-28
Sorta solve #697

### LINKED ISSUE #697: Support for AggregateError syntax highlighting.
It seems [AggregateError](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/AggregateError) is not supported for syntax highlighting. Maybe it should be.

icecream17:
There are *plenty* of unsupported objects: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects, this just adds one

Kinda fixes #697. But tree-sitter still doesn't support

### Description of the Change

Add AggregateError

### Alternate Designs

There seems to be a vertical column alignment, so it kept it.

### Benefits

Slightly more updated code

### Applicable Issues

#697

### DIFF
--- a/grammars/javascript.cson
+++ b/grammars/javascript.cson
@@ -973,10 +973,10 @@
   }
   {
     'match': '''(?x) (?<!\\$) \\b
-      (Array|ArrayBuffer|Atomics|Boolean|DataView|Date|Error|EvalError|Float32Array|Float64Array|Function|Generator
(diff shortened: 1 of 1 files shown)


icecream17:
This pr is good, it's just that the windows github actions are failing: https://github.com/atom/language-javascript/issues/696

PR #709, opened 2021-10-25
Add AggregateError to non-tree-sitter classes

### LINKED ISSUE #697: Support for AggregateError syntax highlighting.

icecream17:
This is a copy of #698, but remade so that there's only 1 commit and so that the tests are reflected as passing.

Copied from #698:
_________________

There are *plenty* of unsupported objects: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects, this just adds one

Kinda fixes #697. But tree-sitter still doesn't support

### Description of the Change

Add AggregateError

### Alternate Designs

There seems to be a vertical column alignment, so it kept it.

### Benefits

Slightly more updated code

### Applicable Issues

#697

### DIFF
--- a/grammars/javascript.cson
+++ b/grammars/javascript.cson
@@ -973,10 +973,10 @@
   }
   {
     'match': '''(?x) (?<!\\$) \\b
-      (Array|ArrayBuffer|Atomics|Boolean|DataView|Date|Error|EvalError|Float32Array|Float64Array|Function|Generator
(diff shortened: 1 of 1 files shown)


darangi:
Thanks for the contribution @icecream17 🙇🏾

PR #714, opened 2022-03-19
use let instead of var to define loop variables

MichaelKreil:
### Requirements

### Description of the Change

Use "let" instead of "var" to define loop variables.

### Alternate Designs

I can't see any other approaches.

### Benefits

Defining loop variables with "var" can lead to unwanted side effects. Therefore it is recommended to use "let".

### Possible Drawbacks

There might be some developers who are not using ES6, like very old legacy versions of node.js. I believe node.js uses "let" since version 4.0.

### Applicable Issues

At some point, we will have to switch to "let". The question is: when, if not now?

### DIFF
--- a/snippets/language-javascript.cson
+++ b/snippets/language-javascript.cson
@@ -28,13 +28,13 @@
     'body': 'else if (${1:true}) {\n\t$2\n}'
   'for':
     'prefix' : 'for'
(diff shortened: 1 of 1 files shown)


icecream17 (approved):

icecream17:
Actually I redact that approval since there's one that also uses const, see https://github.com/atom/language-javascript/pull/691

